PR #90, opened 2015-02-09
Remove the need to do so many manual steps on 'phonegap plugin add'

macdonst:
Remove need to manually delete custom_rules.xml
Copy resources for iOS
Execute hook on build and run
Don't try to unlink a file that doesn't exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,8 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
         <preference name="com.urbanairship.gcm_sender" value="GCM_SENDER_ID" />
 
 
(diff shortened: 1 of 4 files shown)


rlepinski:
This looks great. Could you sign http://urbanairship.com/legal/contribution-agreement/ ?

rlepinski:
Does this require a certain phonegap version? I tried it out locally and the custom_rules.xml still exists and the iOS string file did not get copied.

macdonst:
@rlepinski I am using Cordova/PhoneGap version 4.1.2. It relies on the hooks being run properly. Let me know what version it needs to support and I will try to backport the fix if possible.

rlepinski:
We can up the min version. Ill update to the latest Phonegap and give it another go.

rlepinski:
Support for the plugman hooks was added in 4.0 - https://issues.apache.org/jira/browse/CB-6481

I updated my cordova and it seems to work now. Makes the plugman much easier to use. Thanks!

macdonst:
No problem, happy to help.

PR #91, opened 2015-02-12
Changed engine version to >= 3.4.0, 4.0.0 doesn't exist yet

msangui:
4.0.0 is the cli version not the engine version.
Latest engine version is 3.7


rlepinski:
Fixed in 2.8.1

PR #94, opened 2015-04-23
Update plugin path to urbanairship-cordova

danielcompton:
The path to the plugin has changed, and errors happen when you try to build or add the iOS platform.

screenshot of screenfloat 24-04-15 11 03 51 am

I really appreciate that you work on and maintain the plugin, but its a bit frustrating to run into such a trivial bug.

### DIFF
--- a/scripts/copy_resources.js
+++ b/scripts/copy_resources.js
@@ -18,7 +18,7 @@ module.exports = function (context) {
     ///////////////////////////
         ios : [
             {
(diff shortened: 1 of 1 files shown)


rlepinski:
Its unfortunate that we need a copy resources hook, and it can be so easily broke with a simple ID change. We have an internal jenkins job that should catch this type of thing. Ill will investigate it to make sure this doesnt happen again.

rlepinski:
Thanks for fixing this. It is very much appreciated. :smile:

danielcompton:
No problems, having such quick responses makes things a lot easier :+1:

PR #95, opened 2015-04-23
Add --save command to plugin install

danielcompton:
I'm not sure on your opinion on this, and maybe its a matter of taste, but I prefer to run cordova commands with `--save` to save the information to config.


rlepinski:
I would rather keep our commands as simple as possible, also it matches what they find in cordova's docs - https://cordova.apache.org/docs/en/5.0.0/guide_cli_index.md.html

PR #96, opened 2015-04-28
Add instructions to retrieve the message list

danielcompton:
How can I retrieve the Urban Airship message inbox from the Cordova plugin? I'm not sure how to retrieve the password to make the HTTP call myself, so I'm a little stuck.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,3 +542,5 @@ This event is triggered when your application receives a registration response f
         } 
     });
     
+
+### Retriving the message list


rlepinski:
The plugin doesn't support it currently. 


The easiest approach would be to include the native inbox and message view from the inbox sample and provide a way to show the inbox from the plugin.  The downside here is you will have native code to maintain and native elements to theme. If you are familiar with iOS and Android this might be the best approach for you. 

If you are wanting to do it all within js/html then its probably best to just expose everything through the plugin instead of hitting the APIs directly. For example to get the inbox messages on iOS one could add this to our plugin:

```
- (void)getInboxMessages:(CDVInvokedUrlCommand*)command {
    [self performCallbackWithCommand:command withBlock:^(NSArray *args){
        NSMutableArray *messages = [NSMutableArray array];

        for (UAInboxMessage *message in [UAirship inbox].messageList.messages) {
            NSDictionary *messageInfo = @{ @"title": message.title ?: @"",
                                           @"id": message.messageID,
                                           @"messageBodyUrl": message.messageBodyURL};

            [messages addObject:messageInfo];
        }

        return messages;
    }];
}

- (void)getInboxUser:(CDVInvokedUrlCommand*)command {
    [self performCallbackWithCommand:command withBlock:^(NSArray *args){
        if ([UAirship inboxUser].isCreated) {
            return @{ @"user": [UAirship inboxUser].username, @"password": [UAirship inboxUser].password };
        }

        return @{@"error": @"User not yet created"};
    }];
}

```
You would also need to expose an event when the inbox is updated, a way to mark a message deleted, and a way to mark a message as read.

The problem with this approach is displaying the inbox message. In order to take advantage of our Action framework we need to have the UAWebView delegate available. However I am not sure how one would do that with Cordova.

If you want us to support this feature the best way is to request it through support. Please email support@urbanairship.com.

rlepinski:
Just released 5.0.0 that has message center support.

Display a native Message Center:
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.displayMessageCenter

Methods to provide your own listing:
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.refreshInbox
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.markInboxMessageRead
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.deleteInboxMessage
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#~event:%2522urbanairship.inbox_updated%2522

When providing your own listing, message bodies still need to be displayed in a different webview. Use one of the following methods:
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.displayInboxMessage
- http://docs.urbanairship.com/reference/libraries/urbanairship-cordova/latest/module-UrbanAirship.html#.overlayInboxMessage

PR #98, opened 2015-05-22
Added support to switch on/off Analytics and Location Services

sarriaroman:
This pull request is intended to add a useful feature for those who are not using Location Services or the Analytics service.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,10 @@
 ==============================
+Version 3.1.4 - May 22, 2015
+==============================
+ - The setting "com.urbanairship.enable_analytics_onlaunch" was added to enable/disable Analytics on Launch
+ - The setting "com.urbanairship.enable_location_onlaunch" was added to enable/disable the Location Services on Launch
+
+==============================
 Version 3.1.3 - April 29, 2015
 ==============================
  - Updated iOS Urban Airship SDK to 6.0.2
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,10 +1,4 @@
 ==============================
-Version 3.1.4 - May 22, 2015
-==============================
- - The setting "com.urbanairship.enable_analytics" was added to disable the Analytics Feature
- - The setting "com.urbanairship.enable_location_onlaunch" was added to enable the Location Services on Launch
-
-==============================
 Version 3.1.3 - April 29, 2015
 ==============================
  - Updated iOS Urban Airship SDK to 6.0.2
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,8 @@
 ==============================
 Version 3.1.4 - May 22, 2015
 ==============================
- - The setting "com.urbanairship.enable_analytics_onlaunch" was added to enable/disable Analytics on Launch
- - The setting "com.urbanairship.enable_location_onlaunch" was added to enable/disable the Location Services on Launch
+ - The setting "com.urbanairship.enable_analytics" was added to disable the Analytics Feature
+ - The setting "com.urbanairship.enable_location_onlaunch" was added to enable the Location Services on Launch
 
 ==============================
 Version 3.1.3 - April 29, 2015
--- a/README.md
+++ b/README.md
@@ -47,9 +47,6 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
 		
         <!-- Disable Analytics when the application launches -->
         <preference name="com.urbanairship.enable_analytics" value="true | false" />
-		
-        <!-- Enable Location Services when the application launches -->
-        <preference name="com.urbanairship.enable_location_onlaunch" value="true | false" />
 
         <!-- Override the Android notification icon -->
         <preference name="com.urbanairship.notification_icon" value="ic_notification" />
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
         <!-- Enable push when the application launches -->
         <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
 		
-		<!-- Enable Analytics when the application launches -->
-        <preference name="com.urbanairship.enable_analytics_onlaunch" value="true | false" />
+		<!-- Disable Analytics when the application launches -->
+        <preference name="com.urbanairship.enable_analytics" value="true | false" />
 		
 		<!-- Enable Location Services when the application launches -->
         <preference name="com.urbanairship.enable_location_onlaunch" value="true | false" />
--- a/README.md
+++ b/README.md
@@ -45,10 +45,10 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
         <!-- Enable push when the application launches -->
         <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
 		
-		<!-- Disable Analytics when the application launches -->
+        <!-- Disable Analytics when the application launches -->
         <preference name="com.urbanairship.enable_analytics" value="true | false" />
 		
-		<!-- Enable Location Services when the application launches -->
+        <!-- Enable Location Services when the application launches -->
         <preference name="com.urbanairship.enable_location_onlaunch" value="true | false" />
 
         <!-- Override the Android notification icon -->
--- a/README.md
+++ b/README.md
@@ -44,6 +44,12 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
 
         <!-- Enable push when the application launches -->
         <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
+		
+		<!-- Enable Analytics when the application launches -->
+        <preference name="com.urbanairship.enable_analytics_onlaunch" value="true | false" />
+		
+		<!-- Enable Location Services when the application launches -->
+        <preference name="com.urbanairship.enable_location_onlaunch" value="true | false" />
 
         <!-- Override the Android notification icon -->
         <preference name="com.urbanairship.notification_icon" value="ic_notification" />
--- a/src/android/CordovaAutopilot.java
+++ b/src/android/CordovaAutopilot.java
@@ -55,6 +55,8 @@ public class CordovaAutopilot extends Autopilot {
     static final String ENABLE_PUSH_ONLAUNCH = "com.urbanairship.enable_push_onlaunch";
     static final String NOTIFICATION_ICON = "com.urbanairship.notification_icon";
     static final String NOTIFICATION_ACCENT_COLOR = "com.urbanairship.notification_accent_color";
+	static final String ENABLE_ANALYTICS_ONLAUNCH = "com.urbanairship.enable_analytics_onlaunch";
+	static final String ENABLE_LOCATION_ONLAUNCH = "com.urbanairship.enable_location_onlaunch";
 
     private PluginConfig pluginConfig;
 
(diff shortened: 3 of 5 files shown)


rlepinski:
What advantages does this provide over enabling/disabling analytics and location from JS?

sarriaroman:
- Analytics: when the app is already using another service and the analytics are enabled produce a lot fo overhead that delays the applications startup. So this behaviour brings better experience in this kind of scenarios for the user.
- Location Service: Prevents the unnecessary tracking of the user when the application not require this kind behaviour, some time in some application the users don't want to be tracked or just for a development/company decision.

I hope that the lines explains the why of the features.

rlepinski:
Disabling analytics will have a very insignificant delay on start up time. Analytic event storing and sending happens on a background thread. Location is disabled by default, so the only way to start tracking the user is if you enabled it at the beginning.

Both of these things can be accomplished with running:
```
UAirship.setLocationEnabled(false);
UAirship.setAnalyticsEnabled(false);
```
when cordova is ready.

sarriaroman:
At least on iOS Location starts with the plugin:

```
[[UAirship shared].locationService startReportingSignificantLocationChanges];
````

In the case of Analytics I agree with the insignificant delay when works just with Analytics of UA but when you are using some another one like Google Analytics both are trying to create or read databases, record events, sending data and that's when the problem begins. Also setting the features from the configuration file is easiest to be handled giving the ability to avoid 2 more lines in the case that you need to disable both or maybe in the case when the application is developed by different people in a tem.

**Comment by rlepinski** [comment]
Ok, the UA location services on iOS are a bit tricky. Calling startReportingSignificantLocationChanges doesn't actually start it if location is disabled. I believe nothing will actually be tracked until you call enabled, but this is rather confusing To make it more clear, we should change it to first check if its enabled:
```
if ([UALocationService airshipLocationServiceEnabled]) {
   [[UAirship shared].locationService startReportingSignificantLocationChanges];
}
```

File: src/android/CordovaAutopilot.java [hunk 1]
```diff
@@ -86,6 +88,18 @@ public void onAirshipReady(UAirship airship) {
         if (enablePushOnLaunch) {
             airship.getPushManager().setUserNotificationsEnabled(enablePushOnLaunch);
         }
+		
+		// Enable or disable Analytics on Launch
+		final boolean enableAnalyticsOnLaunch = pluginConfig.getBoolean(ENABLE_ANALYTICS_ONLAUNCH, false);
```
rlepinski: This will only enable it on launch, but thats the default. I think what you are after is the ability to disable it globally, but only when its set in the config. Look at http://docs.urbanairship.com/reference/libraries/android/latest/reference/com/urbanairship/AirshipConfigOptions.html#analyticsEnabled

src/ios/UAirshipPlugin.m
```diff
@@ -72,6 +75,9 @@ - (void)pluginInitialize {
     [UAirship takeOff:config];
 
     [UAirship push].userPushNotificationsEnabledByDefault = [settings[EnablePushOnLaunchConfigKey] boolValue];
+	
+	// Analytics
+	[UAirship shared].analytics.enabled = [settings[EnableAnalyticsOnLaunchConfigKey] boolValue];
```
rlepinski: This is slightly different then android. It will always disabled it unless you set enable_location_onlaunch to true. This will break existing behavior for people. I think you want to do the same thing I suggested for Android.

src/android/CordovaAutopilot.java
rlepinski: Maybe something like:

```
 options.analyticsEnabled = pluginConfig.getBoolean(ENABLE_ANALYTICS, true);
```

And change the flag to just enable analytics

sarriaroman:
I'm doing the changes based on the recomendations.

- Analytics will continue enabled by default.

On Android:
```
options.analyticsEnabled = pluginConfig.getBoolean(ENABLE_ANALYTICS, true);
```

On iOS:
```
if( settings[EnableAnalyticsConfigKey] ) {
	[UAirship shared].analytics.enabled = [settings[EnableAnalyticsConfigKey] boolValue];
}
```

- Location on iOS will be checked as you proposed.

rlepinski:
iOS also has a flag to disable globally - http://docs.urbanairship.com/reference/libraries/ios/latest/Classes/UAConfig.html#//api/name/analyticsEnabled

sarriaroman:
Yes, thanks, I saw the reference after publish my comment.

sarriaroman:
The changes are done based on what we talked. Thanks for your useful time.

CHANGELOG
```diff
@@ -1,4 +1,10 @@
 ==============================
+Version 3.1.4 - May 22, 2015
+==============================
+ - The setting "com.urbanairship.enable_analytics" was added to disable the Analytics Feature
+ - The setting "com.urbanairship.enable_location_onlaunch" was added to enable the Location Services on Launch
```
rlepinski: This is no longer relevant. Actually could you remove the changelog and the package.json and Ill create a release after I have had time to test it?

README.md
```diff
@@ -44,6 +44,12 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
 
         <!-- Enable push when the application launches -->
         <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
+		
+		<!-- Disable Analytics when the application launches -->
+        <preference name="com.urbanairship.enable_analytics" value="true | false" />
+		
+		<!-- Enable Location Services when the application launches -->
+        <preference name="com.urbanairship.enable_location_onlaunch" value="true | false" />
```
rlepinski: No longer relevant

README.md
```diff
@@ -44,6 +44,12 @@ A migration guide for newer releases of the plugin can be found [here](MIGRATION
 
         <!-- Enable push when the application launches -->
         <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
+		
+		<!-- Disable Analytics when the application launches -->
```
rlepinski: Spacing

src/android/CordovaAutopilot.java
```diff
@@ -53,8 +53,12 @@
     static final String IN_PRODUCTION = "com.urbanairship.in_production";
     static final String GCM_SENDER = "com.urbanairship.gcm_sender";
     static final String ENABLE_PUSH_ONLAUNCH = "com.urbanairship.enable_push_onlaunch";
+	static final String ENABLE_LOCATION_ONLAUNCH = "com.urbanairship.enable_location_onlaunch";
```
rlepinski: spacing

src/android/CordovaAutopilot.java
```diff
@@ -53,8 +53,12 @@
...
     static final String NOTIFICATION_ICON = "com.urbanairship.notification_icon";
     static final String NOTIFICATION_ACCENT_COLOR = "com.urbanairship.notification_accent_color";
+	
+	// Enable/Disable features
+	static final String ENABLE_ANALYTICS = "com.urbanairship.enable_analytics";
```
rlepinski: spacing

src/android/CordovaAutopilot.java
```diff
@@ -70,6 +74,9 @@ public AirshipConfigOptions createAirshipConfigOptions(Context context) {
         options.developmentAppSecret = pluginConfig.getString(DEVELOPMENT_SECRET, options.developmentAppSecret);
         options.gcmSender = pluginConfig.getString(GCM_SENDER, options.gcmSender);
         options.inProduction = pluginConfig.getBoolean(IN_PRODUCTION, options.inProduction);
+		
+		// Analytics. Enabled by Default
+		options.analyticsEnabled = pluginConfig.getBoolean(ENABLE_ANALYTICS, true);
```
rlepinski: spacing

src/android/CordovaAutopilot.java
```diff
@@ -86,6 +93,12 @@ public void onAirshipReady(UAirship airship) {
         if (enablePushOnLaunch) {
             airship.getPushManager().setUserNotificationsEnabled(enablePushOnLaunch);
         }
+		
+		// Enable or disable Location Services on Launch
+		final boolean enableLocationOnLaunch = pluginConfig.getBoolean(ENABLE_LOCATION_ONLAUNCH, false);
+        if (enableLocationOnLaunch) {
+            airship.getLocationManager().setLocationUpdatesEnabled(enableLocationOnLaunch);
+        }
```
rlepinski: lets remove this

src/ios/UAirshipPlugin.m
```diff
@@ -51,6 +51,9 @@ @implementation UAirshipPlugin
 NSString *const ProductionConfigKey = @"com.urbanairship.in_production";
 NSString *const EnablePushOnLaunchConfigKey = @"com.urbanairship.enable_push_onlaunch";
 NSString *const ClearBadgeOnLaunchConfigKey = @"com.urbanairship.clear_badge_onlaunch";
+NSString *const EnableLocationOnLaunchConfigKey = @"com.urbanairship.enable_location_onlaunch";
```
rlepinski: remove this

src/ios/UAirshipPlugin.m
```diff
@@ -65,7 +68,11 @@ - (void)pluginInitialize {
     config.inProduction = [settings[ProductionConfigKey] boolValue];
     config.developmentLogLevel = UALogLevelTrace;
     config.productionLogLevel = UALogLevelTrace;
-
+	
+	// Analytics. Enabled by Default
+	if( settings[EnableAnalyticsConfigKey] != nil ) {
+		config.analyticsEnabled = [settings[EnableAnalyticsConfigKey] boolValue];
```
rlepinski: spacing is a little off.

```
if (settings[EnableAnalyticsConfigKey] != nil) {
     config.analyticsEnabled = [settings[EnableAnalyticsConfigKey] boolValue];
}
```

src/ios/UAirshipPlugin.m
```diff
@@ -79,8 +86,15 @@ - (void)pluginInitialize {
 
     [UAirship push].pushNotificationDelegate = self;
     [UAirship push].registrationDelegate = self;
-
-    [[UAirship shared].locationService startReportingSignificantLocationChanges];
+	
+	// Enable Location Service. Disabled by Default
+	if( settings[EnableLocationOnLaunchConfigKey] ) {
```
rlepinski: Remove this

src/ios/UAirshipPlugin.m
```diff
@@ -79,8 +86,15 @@ - (void)pluginInitialize {
...
+		[UALocationService setAirshipLocationServiceEnabled:[settings[EnableLocationOnLaunchConfigKey] boolValue]];
+	}
+
+	if ([UALocationService airshipLocationServiceEnabled]) {
+   		[[UAirship shared].locationService startReportingSignificantLocationChanges];
+    }
```
rlepinski: spacing

rlepinski:
It mostly looks fine. I think the location stuff needs to be removed because its disabled by default. The code has inconsistent style that needs to be fixed. 

Its a long weekend, I will test this again on Tuesday and hopefully do a release. Thanks for all the effort!

sarriaroman:
Regarding the location stuff will give the Cordova developer more control and avoid to depend on the WebApplication developer if not the same. But if you want I can remove it. This case is based on my own experience working with the plugin, sometimes I just need to use the same code for different webapps code and avoid that the webapp developer forget to enable the feature.

I will fix the styling, I don't know why was changed, sorry about that.

rlepinski:
The location stuff encourages bad practices. On iOS, it will prompt the user to accept the use of location. If you are doing this on app launch, it has a higher chance it will be declined. Push is the same way, and the only reason the flag exists for push is to support old behavior people might of relied on.

sarriaroman:
I need to agree with the bad practice part. I will change it. Thanks

src/ios/UAirshipPlugin.m
```diff
@@ -65,7 +67,11 @@ - (void)pluginInitialize {
     config.inProduction = [settings[ProductionConfigKey] boolValue];
     config.developmentLogLevel = UALogLevelTrace;
     config.productionLogLevel = UALogLevelTrace;
-
+    
+    // Analytics. Enabled by Default
+    if( settings[EnableAnalyticsConfigKey] != nil ) {
```
rlepinski: minor style change `if (settings[EnableAnalyticsConfigKey] != nil) {`

rlepinski:
Everyone looks great except for the one minor style change.

Before I can merge this work, I need you to fill out a contribution agreement -  http://urbanairship.com/legal/contribution-agreement/

Thanks!

src/ios/UAirshipPlugin.m
```diff
@@ -65,7 +67,11 @@ - (void)pluginInitialize {
     config.inProduction = [settings[ProductionConfigKey] boolValue];
     config.developmentLogLevel = UALogLevelTrace;
     config.productionLogLevel = UALogLevelTrace;
-
+    
+    // Analytics. Enabled by Default
+    if(settings[EnableAnalyticsConfigKey] != nil) {
```
rlepinski: still missing the space `if(` -> `if (`

sarriaroman:
I think is done now. Sorry about the styling. I already filled the agreement.

PR #99, opened 2015-08-03
Sending Activity context to PlayServicesUtils#handleAnyPlayServicesError

krebernisak:
When starting new activity using the application context, Android app will crash if the start intent does not have FLAG_ACTIVITY_NEW_TASK flag set.


rlepinski:
I am confused on how your change adds the FLAG_ACTIVITY_NEW_TASK flag? It looks like its just changing what context you are using. If you are getting a NPE, then its because airship did not takeoff properly and UAirship.getApplicationContext() is returning null.

rlepinski:
Looks good!
Could you fill out - http://urbanairship.com/legal/contribution-agreement ?

krebernisak:
Yes, I've filled out the contribution agreement.

rlepinski:
Thanks

PR #100, opened 2015-08-25
Return device token from UARegistrationDelegate registrationSucceededForChannelID

kurtisf:
Return the device token along with the channel ID from the urbanairship.registration event.  This is helpful to be able to send the device token to the back end server to be used with the UA API.

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -593,7 +593,13 @@ NSString *const EnableAnalyticsConfigKey = @"com.urbanairship.enable_analytics";
     UA_LINFO(@"Channel registration successful %@.", channelID);
 
     if (self.registrationCallbackID) {
-        NSDictionary *data = @{ @"channelID":channelID };
+        NSDictionary *data;
+        if (deviceToken == nil) {
+            data = @{ @"channelID":channelID };
+        }
+        else {
+            data = @{ @"channelID":channelID, @"deviceToken":deviceToken };
+        }
         CDVPluginResult* result = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK messageAsDictionary:data];
         [result setKeepCallbackAsBool:YES];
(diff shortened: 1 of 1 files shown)


rlepinski:
Why are you using the device token instead of the channel ID?

kurtisf:
It is an existing API/Service implementation that already uses the device
token directly from iOS in other applications.

On Tue, Aug 25, 2015 at 4:02 PM, Ryan Lepinski <notifications@github.com>
wrote:

> Why are you using the device token instead of the channel ID?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/urbanairship/phonegap-ua-push/pull/100#issuecomment-134765541>
> .
>

-- 
This email may contain confidential information for the sole use of the 
intended recipient(s). If you are not an intended recipient, please notify 
the sender and delete all copies immediately.

src/ios/UAirshipPlugin.m
```diff
@@ -593,7 +593,13 @@ - (void)registrationSucceededForChannelID:(NSString *)channelID deviceToken:(NSS
     UA_LINFO(@"Channel registration successful %@.", channelID);
 
     if (self.registrationCallbackID) {
-        NSDictionary *data = @{ @"channelID":channelID };
+        NSDictionary *data;
+        if (deviceToken == nil) {
```
rlepinski: we use `if (!deviceToken)` instead of direct nil comparison. However in this case you should reverse your logic and do `if (deviceToken)`

src/ios/UAirshipPlugin.m
```diff
@@ -593,7 +593,13 @@ - (void)registrationSucceededForChannelID:(NSString *)channelID deviceToken:(NSS
...
+            data = @{ @"channelID":channelID };
+        }
```
rlepinski: `} else {`

rlepinski:
Looks good. Could you also fill out http://urbanairship.com/legal/contribution-agreement/ ?

Thanks

rlepinski:
Awesome. Thanks!

PR #102, opened 2015-11-17
Fix config parser for 4.5.0 and Cordova-commons

j-mcnally:
Config parser is totally broke with latest cordova. this allows it to find it by module rather than relative path.

### DIFF
--- a/scripts/copy_resources.js
+++ b/scripts/copy_resources.js
@@ -3,11 +3,11 @@
 var fs = require('fs');
 var path = require('path');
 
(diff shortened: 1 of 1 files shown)


rlepinski:
Yep. We are actually working on fixing this.

This is slightly better than our fix. We just updated the path. Where you able to find docs on the proper way of getting the module?

Could you fill out http://docs.urbanairship.com/contribution-agreement.html ? I need it before I can merge.

Thanks!

rlepinski:
Thank you!

PR #104, opened 2016-05-18
[MB-1787] do UI stuff on the main queue, gentler logging

marc-scig:
We need to be careful about not making UI-related calls from the background, which is the default now for our plugin callbacks. Much of UIKit is not thread safe, leading to undefined behavior or this gnarly gem as reported by @linkmh:

```
bool _WebTryThreadLock(bool), 0x160805c00: Tried to obtain the web lock from a thread other than the main thread or the web thread. This may be a result of calling to UIKit from a secondary thread. Crashing now...
```

While we're at it, the log message about missing listeners is probably TMI for most cases so I bumped the log level up to trace in case it ever comes in handy.

I also added a bit to the plugin result logic that keeps it from logging an error when plugin calls don't return a value, which happens quite a lot now.


rlepinski:
🤓

marc-scig:
Display MC button isn't working, I must be doing something wrong. Hold up.

PR #120, opened 2016-08-11
Fire a deep link event when deep link action is called

gatlinhebert:
Enables the ability to handle deep links triggered from in-app messages, landing pages, and the message center.

### DIFF
--- a/Example/index.html
+++ b/Example/index.html
@@ -45,20 +45,20 @@
           }
         }
 
+        //Deep link callback
+        var handleDeepLink = function(event)  {
+            console.log("Deep link: " + event.deepLink)
+        }
+
         // Register for any urban airship events
         document.addEventListener("urbanairship.registration", onRegistration, false)
         document.addEventListener("urbanairship.push", onPushReceived, false)
+        document.addEventListener("urbanairship.deep_link", handleDeepLink, false)
         
         // Handle resume
         document.addEventListener("resume", function() {
           console.log("Device resume!")
           
-          UAirship.getDeepLink(true, function(deepLink) {
-            if (deepLink) {
-              console.log("Deep link: " + deepLink)
-            }
-          })
-
           UAirship.resetBadge()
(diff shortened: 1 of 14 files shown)


rlepinski:
Does the event get delivered when starting the app from a notification that contains a deep link?

gatlinhebert:
It fails when launching from a notification as the listener callback is unavailable.

rlepinski:
We probably need to add a map of pending events if the listener is not there, then fire all those events in the map as soon as the listener is available.

Example/index.html
```diff
@@ -65,6 +73,7 @@
           // Reregister for urbanairship events if they were removed in pause event
           document.addEventListener("urbanairship.registration", onRegistration, false)
           document.addEventListener("urbanairship.push", onPushReceived, false)
+          document.addEventListener("urbanairship.deeplink", handleDeepLink, false)
```
marc-scig: the one above has an underscore, is that intentional?

rlepinski: nope, good catch

Example/index.html
```diff
@@ -74,6 +83,7 @@
           // Remove urbanairship events.  Important on android to not receive push in the background.
           document.removeEventListener("urbanairship.registration", onRegistration, false)
           document.removeEventListener("urbanairship.push", onPushReceived, false)
+          document.removeEventListener("urbanairship.deeplink", handleDeepLink, false)
```
marc-scig: samesies

marc-scig:
🍻  🐻  🐗  📊  🍔

Example/index.html
```diff
@@ -45,9 +45,17 @@
           }
         }
 
+        //Deep link callback
+        var handleDeepLink = function(event)  {
+          if (event.value) {
```
gatlinhebert: if (event.deepLink)

gatlinhebert:
+1 🎉

PR #128, opened 2016-09-01
[MB-1916] Release 5.4.0

hcrowell:
Release 5.4.0:
 - Updated iOS Urban Airship SDK to 7.3.0
 - Updated Android Urban Airship SDK to 7.3.0
 - All events are now fired after the 'deviceready' event
 - Added document event 'urbanairship.deep_link' when a new deep link is received
 - Added `isAppNotificationsEnabled` to check if the user opted in to notifications for the device
 - Added `setAssociatedIdentifier` to set device identifiers for Urban Airship Connect

Testing:
* Received pushes on iOS 9.3.5 iPod and Android  6.0.1 nexus5x


src/ios/UAirshipPlugin.h
```diff
@@ -277,6 +277,7 @@
  * Records the current location.
  *
  * @param command The cordova command.
+ * @deprecated since 5.4.0
  */
 - (void)recordCurrentLocation:(CDVInvokedUrlCommand *)command;
```
hcrowell: I deprecated this method since we're no longer using the old `UALocationService`, but wasn't sure about removing it.

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+============================
+Version 5.4.0 - Sept 1, 2016
+============================
+ - Updated iOS Urban Airship SDK to 7.3.0
+ - Updated Android Urban Airship SDK to 7.3.0
+ - All events are now fired after the 'deviceready' event
+ - Added document event 'urbanairship.deep_link' when a new deep link is received
+ - Added `isAppNotificationsEnabled` to check if the user opted in to notifications for the device
+ - Added `setAssociatedIdentifier` to set device identifiers for connect
```
z0br0wn: should connect be capitalized?

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+============================
+Version 5.4.0 - Sept 1, 2016
+============================
+ - Updated iOS Urban Airship SDK to 7.3.0
+ - Updated Android Urban Airship SDK to 7.3.0
+ - All events are now fired after the 'deviceready' event
+ - Added document event 'urbanairship.deep_link' when a new deep link is received
```
z0br0wn: whats a document event?

z0br0wn:
otherwise 😎 (granted i'm not very familiar with the ios location service changes)

CHANGELOG
hcrowell: A document event allows JavaScript to register an event handler on the element in the HTML document. For this case, we're adding the `urbanairship.deep_link` event which we can register for this event with a `handleDeepLink` callback.

marc-scig:
🎱

PR #131, opened 2016-09-16
[MB-1962] Set notification large icon.

hcrowell:
Allows configuring the plugin to use a custom notification large icon.

// config.xml
`<!-- Override the Android notification large icon -->`
`<preference name="com.urbanairship.notification_large_icon" value="ua_ic_close" />`

Example where the large icon is `X`:
marshmallow_6_api_23_phonegap_large_icon_uaclosebutton_closeup

### DIFF
--- a/src/android/CordovaAutopilot.java
+++ b/src/android/CordovaAutopilot.java
@@ -119,7 +119,7 @@ public class CordovaAutopilot extends Autopilot {
         String notificationIconName = pluginConfig.getString(NOTIFICATION_ICON, null);
         if (!UAStringUtil.isEmpty(notificationIconName)) {
             int id  = context.getResources().getIdentifier(notificationIconName, "drawable", context.getPackageName());
-            if (id > 0) {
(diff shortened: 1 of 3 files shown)


src/android/CordovaAutopilot.java
```diff
@@ -125,6 +126,17 @@ public void onAirshipReady(UAirship airship) {
             }
         }
 
+        // Notification large icon
+        String notificationLargeIconName = pluginConfig.getString(NOTIFICATION_LARGE_ICON, null);
+        if (!UAStringUtil.isEmpty(notificationLargeIconName)) {
+            int id  = context.getResources().getIdentifier(notificationLargeIconName, "drawable", context.getPackageName());
+            if (id > 0) {
```
rlepinski: can we update these to check for `id != 0`
only 0 is an invalid identifier.

hcrowell: sure!

PR #144, opened 2016-10-06
[MB-1976] Added iOS 10 foreground presentation support.

hcrowell:
Added iOS 10 foreground presentation support.

Specify the optional default foreground presentation options for iOS 10 via the config.xml:
```
<!-- iOS 10 alert foreground notification presentation option -->
<preference name="com.urbanairship.ios_foreground_notification_presentation_alert" value="true"/>

<!-- iOS 10 badge foreground notification presentation option -->
<preference name="com.urbanairship.ios_foreground_notification_presentation_badge" value="true"/>

<!-- iOS 10 sound foreground notification presentation option -->
<preference name="com.urbanairship.ios_foreground_notification_presentation_sound" value="true"/>
```

Testing:
* Tested foreground push notifications with various presentation options on iOS 10.0.2 iPod

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -65,6 +65,9 @@ NSString *const EnablePushOnLaunchConfigKey = @"com.urbanairship.enable_push_onl
 NSString *const ClearBadgeOnLaunchConfigKey = @"com.urbanairship.clear_badge_onlaunch";
 NSString *const EnableAnalyticsConfigKey = @"com.urbanairship.enable_analytics";
 NSString *const AutoLaunchMessageCenterKey = @"com.urbanairship.auto_launch_message_center";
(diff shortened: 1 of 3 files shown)


rlepinski:
👍

src/ios/UAirshipPlugin.m
```diff
@@ -115,6 +118,35 @@ - (void)pluginInitialize {
     [UAirship push].registrationDelegate = self;
     [UAirship inbox].delegate = self;
 
+    // Set the iOS default foreground presentation options if specified in the config
+    if (settings[NotificationPresentationAlertKey] != nil ||
```
rlepinski: why do we need to check for any of the 3?

crow:
^with extra checks addressed 🍔

marc-scig:
✌️

PR #150, opened 2016-10-26
Removed support email address from repo

jenniferjsmmiller:
Hi all,

Per [SUPPORT-5868](https://tofurkey.urbanairship.com/browse/SUPPORT-5868):

The support@urbanairship.com email address is listed on the phonegap-ua-push repository on Github. We do not want this listed on a public repository as email support is reserved for paying customers. We direct any free customers to the forums, so I think it makes sense to list http://support.urbanairship.com/ in place of the support email address.

I only modified the README.md and the package.json file. I just removed the bugs field from the package.json. I can add it back in there if need be, and just replace 'email' with 'url' and make the url the support website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ We accept pull requests! If you would like to submit a pull request, please fill
 
 ### Issues
 
-Please contact support@urbanairship.com for any issues integrating or using this plugin.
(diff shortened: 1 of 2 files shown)


rlepinski:
👍  Looks great

PR #157, opened 2016-12-08
Require cordova-cli 6.4.0

hcrowell:
Update readme to require cordova-cli 6.4.0

Note: I think we can ignore the check failure, since this is a one line non-code change in readme file. I believe it is due to the new build machine and will fix it when I get into the office.


madelgi:
🌴

rlepinski:
👍

PR #182, opened 2017-05-23
Cordova Release 6.7.1

BrianBatchelder:
Tested on iOS device and Android emulator.


rlepinski:
👍

CHANGELOG
```diff
@@ -1,4 +1,10 @@
 ============================
+Version 6.8.0 - May 23, 2017
```
madelgi: should this be a patch release since both sdk updates are patch releases?

madelgi:
:+1: so long as the above isn't a problem

PR #192, opened 2017-08-07
[CE-3539] Fix displayMessageCenter on Android

rlepinski:
Checks to make sure the message ID is not null or empty before setting the intent


crow:
🍔

marc-scig:
💯

PR #208, opened 2018-01-27
[MB-2557] Compatibility with Android 7.0.0

rlepinski:
Android 7.0.0 was breaking due to Cordova thinking our plugin was an eclipse style plugin because we had a "res" folder inside our plugin. None of those resources were technically needed, so I removed them and patched the plugin to no longer use them. This biggest change here is the plugin.xml whitespace clean up.


rlepinski:
retest this please

marc-scig:
👨‍🔬

PR #211, opened 2018-01-31
[MB-2531] Add response info to notification event

rlepinski:
Adds notification response info to the open event.

I also fixed a few issues when I was testing. Lots of main thread checker warnings for iOS and build issues for Android. 

Testing:
 - Android registration event, notification opens, etc..
- iOS registration event. Still need to test a interactive notification.

### DIFF
--- a/src/android/events/NotificationOpenedEvent.java
+++ b/src/android/events/NotificationOpenedEvent.java
@@ -25,6 +25,10 @@
 
 package com.urbanairship.cordova.events;
 
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import com.urbanairship.AirshipReceiver;
 import com.urbanairship.Logger;
 import com.urbanairship.push.PushMessage;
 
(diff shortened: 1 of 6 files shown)


src/android/events/NotificationOpenedEvent.java
```diff
@@ -41,17 +45,53 @@
...
+
+
+    @Override
+    public JSONObject getEventData() {
+        JSONObject jsonObject = super.getEventData();
+
+        try {
+            if (actionButtonInfo != null) {
+                jsonObject.put(ACTION_ID, actionButtonInfo.getButtonId());
+                jsonObject.put(IS_FOREGROUND, actionButtonInfo.isForeground());
```
marc-scig: do we need `IS_FOREGROUND` in the event data here? slash where's this stuff spec'ed out?

src/ios/UAirshipPlugin.m
```diff
@@ -237,24 +237,22 @@ - (CDVPluginResult *)pluginResultForValue:(id)value status:(CDVCommandStatus)sta
...
  * @param command The cordova command.
  * @param block The UACordovaExecutionBlock to execute.
  */
 - (void)performCallbackWithCommand:(CDVInvokedUrlCommand *)command withBlock:(UACordovaExecutionBlock)block {
-    [self.commandDelegate runInBackground:^{
-        UACordovaCompletionHandler completionHandler = ^(CDVCommandStatus status, id value) {
-            CDVPluginResult *result = [self pluginResultForValue:value status:status];
-            [self.commandDelegate sendPluginResult:result callbackId:command.callbackId];
-        };
+    UACordovaCompletionHandler completionHandler = ^(CDVCommandStatus status, id value) {
```
marc-scig: is this related to the "main thread checker" bit you were mentioning? does this effectively make all cordova callbacks run on the UI thread?

src/ios/UAirshipPlugin.m
```diff
@@ -1014,4 +1009,40 @@ -(NSInteger)parseLogLevel:(NSString *)logLevel defaultLogLevel:(NSInteger)defaul
     return defaultValue;
 }
 
+
+- (UANotificationAction *)notificationActionForCategory:(NSString *)category actionIdentifier:(NSString *)identifier {
```
marc-scig: this looks like a duplication of the class method in UAAppIntegration. Any chance we could use that instead?

src/ios/UAirshipPlugin.m
```diff
@@ -791,18 +799,14 @@ - (void)showInbox {
 
 - (void)displayMessageCenter:(CDVInvokedUrlCommand *)command {
     [self performCallbackWithCommand:command withBlock:^(NSArray *args, UACordovaCompletionHandler completionHandler) {
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [[UAirship defaultMessageCenter] display];
-        });
+        [[UAirship defaultMessageCenter] display];
```
BrianBatchelder: Why did this have to change? I thought the warning was about doing stuff _off_ the main thread.

src/ios/UAirshipPlugin.m
rlepinski: Yes, but all of ours calls are none blocking

src/ios/UAirshipPlugin.m
rlepinski: We no longer call everything on a background thread

src/ios/UAirshipPlugin.m
rlepinski: We could if it was public. Right now its internal only.

src/android/events/NotificationOpenedEvent.java
rlepinski: IS_FOREGROUND is if the notification response is a foreground or background action.

src/ios/UAirshipPlugin.m
BrianBatchelder: OK.

marc-scig:
🐻

PR #213, opened 2018-02-01
Fix docs

rlepinski:
Fixes docs build. Getting some error due to an incompatible version of jsdocs with the current version of node.


crow:
:hamburger:

marc-scig:
📞

PR #214, opened 2018-02-08
MB-2534 Corova plugin 7.0

marc-scig:
This ended up being a little more involved than anticipated, since a fair amount has changed at the interface level in SDK 9. But I think I got everything.

I ended up removing the setDisplayASAPEnabled stuff, since it's no longer a thing, and fixing up the iOS plugin source code to reflect our new, ex-default message center reality. Android gradle references should be up to date. Tested both iOS and android sample projects, though not super extensively. I could probably use some help making sure everything checks out.


rlepinski:
👍 
Thanks for breaking out the iOS SDK update into its own commit.

rlepinski:
update to 9.0.2 on iOS

rlepinski:
👍

rlepinski:
👍

PR #218, opened 2018-04-05
update getEventData to handle integer value

leegenes:
Android's PushEvent is throwing an exception on `getEventData` while looping through `pushBundle`. We're passing `google.ttl` as an integer, but calling `getString` on it.

This gets the value as and integer and converts to a string before adding to `extras`.


rlepinski:
Its not throwing exception right? Its just logging a gross error?

leegenes:
@rlepinski yeah. that's true. my bad. 
and is putting null to extras rather than the integer passed as a string.

rlepinski:
👍

crow:
🍔

PR #227, opened 2018-05-30
Addresses https://github.com/urbanairship/urbanairship-cordova/issues…

chrisweight:
Addresses https://github.com/urbanairship/urbanairship-cordova/issues/224

Adds alert body, title (if available) and push ID to 'extras' stanza
Uses same keys as Android for consistency

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -977,7 +977,7 @@ NSString *const ExtrasPushId = @"com.urbanairship.push.PUSH_ID";
         [extras removeObjectForKey:@"aps"];
     }
 
-    terif([[extras allKeys] containsObject:@"_"]) {
+    if([[extras allKeys] containsObject:@"_"]) {
         [extras setValue:extras[@"_"] forKey:ExtrasPushId];
         [extras removeObjectForKey:@"_"];
     }
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -974,19 +974,19 @@ NSString *const ExtrasPushId = @"com.urbanairship.push.PUSH_ID";
     NSMutableDictionary *extras = [NSMutableDictionary dictionaryWithDictionary:notificationContent.notificationInfo];
 
     if([[extras allKeys] containsObject:@"aps"]) {
-				[extras removeObjectForKey:@"aps"];
+        [extras removeObjectForKey:@"aps"];
     }
 
-    if([[extras allKeys] containsObject:@"_"]) {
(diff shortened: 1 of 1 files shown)


src/ios/UAirshipPlugin.m
```diff
@@ -65,6 +65,11 @@ @implementation UAirshipPlugin
 NSString *const EventRegistration = @"urbanairship.registration";
 NSString *const EventDeepLink = @"urbanairship.deep_link";
 
+// Extras
+NSString *const ExtrasTitle = @"com.urbanairship.title";
+NSString *const ExtrasAlert = @"com.urbanairship.push.ALERT";
+NSString *const ExtrasPushId = @"com.urbanairship.push.PUSH_ID";
```
rlepinski: Instead of dumping these in the "extras" section, why not the top level?

rlepinski: I am pretty sure on iOS its the `sendId` not the `pushId`.

src/ios/UAirshipPlugin.m
```diff
@@ -969,13 +974,20 @@ - (NSDictionary *)pushEventFromNotification:(UANotificationContent *)notificatio
...
     if([[extras allKeys] containsObject:@"_"]) {
+			  [extras setValue:extras[@"_"] forKey:ExtrasPushId];
         [extras removeObjectForKey:@"_"];
     }
 
+	 	if ([notificationContent respondsToSelector:NSSelectorFromString(@"alertTitle")]) {
+        [extras setValue:notificationContent.alertTitle forKey:ExtrasTitle];
+    }
+
+ 		[extras setValue:notificationContent.alertBody ?: @"" forKey:ExtrasAlert];
```
rlepinski: Formatting is a bit messed up

chrisweight: That's weird...looks fine in VS Code...will dump it into XCode to check.

rlepinski:
@chrisweight Yeah #231 addresses your issue and handles the feedback I had for this PR. It will add `title` and `subtitle` at a top level and include `aps` in case we miss anything in the iOS payload. We will not expose the send ID as its usually not what people are looking for (ID for the individual device send), and push ID is not available for both iOS and Android.

src/ios/UAirshipPlugin.m
rlepinski: These extras are just how Android receives the payload. What I was requesting was adding "title" to the top level for both platforms of the push event instead of relying on these long extra keys.

chrisweight:
All sounds good! Cheers Ryan :)

PR #228, opened 2018-06-07
[MOBILE-193] support for loading custom notification categories

marc-scig:
This lets users supply custom notification categories in plist/xml format as an app resource, so that the plugin can find it at takeOff time and register the automatically. I started out building a JS interface for this, but eventually backed away from that approach because (a) it would require its own persistence mechanism and (b) the category/group model is so different between iOS and Android that it's difficult to make an API representing a common subset that is useful enough to be worthwhile.

Aside from some confusion in their docs about the app structure (https://cordova.apache.org/announcements/2017/12/04/cordova-android-7.0.0.html) this is pretty straightforward. The relevant entry in config.xml looks like this, with the resource files parallel to it at build time

```
<platform name="android">
    <resource-file src="ua_custom_notification_buttons.xml" target="app/src/main/res/xml/ua_custom_notification_buttons.xml" />
</platform>
<platform name="ios">
    <resource-file src="UACustomNotificationCategories.plist" />
</platform>
```

Since the Android button groups parser is package protected, this also adds a little util class I mostly cribbed from ryan's work on 9.3. We may want to hold off on releasing this until 9.3 is out, in which case it might make sense just to fold this into the firebase ticket and do one big update. In that case, the util class will be unnecessary and can go away. 

Edit: updated README with examples


src/android/CordovaAutopilot.java
```diff
@@ -198,13 +200,20 @@ public boolean apply(ActionArguments actionArguments) {
...
             @Override
             public void onInboxUpdated() {
                 UAirshipPluginManager.shared().inboxUpdated();
             }
         });
+
+        String packageName = UAirship.shared().getPackageName();
+        @XmlRes int resId = context.getResources().getIdentifier("ua_custom_notification_buttons", "xml", packageName);
+
+        if (resId > 0) {
```
rlepinski: negative res values are ok, so you want to check != 0

rlepinski:
You should update the README/platform docs about how to setup the categories

README.md
```diff
@@ -87,14 +87,30 @@ the GCM/FCM sender ID either needs to be prefixed with `sender:` or you can disa
...
+       
+       ...
+
+       <!-- Optional: include custom notification categories in plist format -->
+       <platform name="iOS">
+            ...
+            <resource-file src="UACustomNotificationCategories.plist" />
+       </platform>
+
+4. *(Android Only)* Add a reference to your google-servies.json file, for configuring FCM (parallel to config.xml):
```
BrianBatchelder: What does "parallel to config.xml" mean? (I see that it was already there, but perhaps we can make it more clear)

src/android/CordovaAutopilot.java
```diff
@@ -198,13 +199,20 @@ public boolean apply(ActionArguments actionArguments) {
...
             @Override
             public void onInboxUpdated() {
                 UAirshipPluginManager.shared().inboxUpdated();
             }
         });
+
+        String packageName = UAirship.shared().getPackageName();
+        @XmlRes int resId = context.getResources().getIdentifier("ua_custom_notification_buttons", "xml", packageName);
+
+        if (resId != 0) {
```
BrianBatchelder: I'm sure I'm not the only one who hates magic numbers. I _assume_ zero means invalid Id, but, yuch.

rlepinski: 0 is the only value invalid for android resources

rlepinski:
👍

README.md
marc-scig: parallel in terms of file paths

src/android/CordovaAutopilot.java
marc-scig: While we're talking about things we hate, what's up with Java not having unsigned integers, anyway

PR #230, opened 2018-06-09
[ios-updates] update iOS SDK to 9.2

marc-scig:
This was not explicitly asked for but we might as well update the iOS SDK while we're at it. Apologies for the line noise, this is basically all Airship and associated headers. The only real changes are in plugin.xml.


rlepinski:
👍

PR #231, opened 2018-06-11
[MOBILE-200] title and subtitle, aps push event fields

marc-scig:
This is a PR that addresses the same stuff as https://github.com/urbanairship/urbanairship-cordova/pull/227, but with a structure we'd prefer.

* Adds title and subtitle to top level push event object. (subtitle maps to "summary" on android for historical reasons)
* APS dictionary on iOS now goes in its own top-level field as well, in case folks want access to the raw push data for whatever reason

Note: I'm not sure how best to document this change, aside from making a note in the CHANGELOG. Platform extdocs?

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -965,20 +965,32 @@ NSString *const EventDeepLink = @"urbanairship.deep_link";
         return @{ @"message": @"", @"extras": @{}};
     }
 
-    // remove extraneous key/value pairs
-    NSMutableDictionary *extras = [NSMutableDictionary dictionaryWithDictionary:notificationContent.notificationInfo];
+    NSMutableDictionary *info = [NSMutableDictionary dictionaryWithDictionary:notificationContent.notificationInfo];
 
-    if([[extras allKeys] containsObject:@"aps"]) {
-        [extras removeObjectForKey:@"aps"];
-    }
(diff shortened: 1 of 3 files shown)


rlepinski:
@marc-scig Document the payload here - https://github.com/urbanairship/urbanairship-cordova/blob/master/www/UrbanAirship.js#L168

marc-scig:
Payload is doc'ed. The event links still do not work and I don't know why, but at least it's in there now. We can figure out the other bit later.

src/ios/UAirshipPlugin.m
```diff
@@ -965,20 +965,32 @@ - (NSDictionary *)pushEventFromNotification:(UANotificationContent *)notificatio
...
-    result[@"extras"] = extras;
+
+    // Set the title and subtitle as top level objects, if present
+    NSString *title = notificationContent.alertTitle;
+    NSString *subtitle = notificationContent.notification.request.content.subtitle;
+    [result setValue:title forKey:@"title"];
+    [result setValue:subtitle forKey:@"subtitle"];
+
+    // Set the remaining info as extras
+    result[@"extras"] = info;
```
BrianBatchelder: FWIW, I'd move this up to line 982, just after the other usage of `info`. I think it is clearer if the uses of `info` are grouped together.

PR #235, opened 2018-06-14
[MOBILE-215] Enable takeOff from JS for Android

rlepinski:
Enables calling takeOff from JS. PR only covers Android, iOS will get its own PR.

Config is applied with the following rules:
- Values always default to config.xml values
- Values supplied in JS takeOff will override config.xml values
- If takeOff is already called, new values will be applied during the next app init

Testing:
- Testing setting takeOff and setting notification config in JS


package.json
```diff
@@ -42,6 +42,6 @@
     "minami": "1.2.3"
   },
   "scripts": {
-    "generate-docs": "node_modules/.bin/jsdoc ./www/UrbanAirship.js -t ./node_modules/minami/ -r jsdoc_readme.md"
+    "generate-docs": "jsdoc ./www/UrbanAirship.js -t ./node_modules/minami/ -r jsdoc_readme.md"
```
rlepinski: No need to specify the node_modules dir when you define it in the package.json script

src/android/ConfigUtils.java
```diff
@@ -0,0 +1,107 @@
...
+import com.urbanairship.Logger;
+
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Config Utils.
+ */
+class ConfigUtils {
```
rlepinski: Existing code, just moved into a utils class

src/android/UAirshipPlugin.java
```diff
@@ -97,46 +66,76 @@
...
+        if (!isAirshipAction && !isGlobalAction) {
             Logger.debug("Invalid action: " + action);
             return false;
         }
 
         executorService.execute(new Runnable() {
             @Override
             public void run() {
+
+                if (!pluginManager.isAirshipAvailable() && isAirshipAction) {
```
rlepinski: This will just error on any calls from the JS layer if takeOff is not called.

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+        return getConfigBoolean(AUTO_LAUNCH_MESSAGE_CENTER, true);
+    }
+
+    /**
+     * Gets the notification sound.
+     *
+     * @return The notification sound.
+     */
+    @Nullable
+    public Uri getNotificationSound() {
```
BrianBatchelder: Do you need a setter for the sound?

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * Gets the enable push on launch option.
+     *
+     * @return {@code true} to enable push on launch, otherwise {@code false}.
+     */
+    public boolean getEnablePushOnLaunch() {
```
BrianBatchelder: Do you need a setter for push on launch?

rlepinski: We only have setters for things that we support setting.

BrianBatchelder: :thumbsup:

src/android/PluginManager.java
BrianBatchelder: :thumbsup:

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+            return defaultValue;
+        }
+
+        return Boolean.parseBoolean(value);
+    }
+
+    /**
+     * Gets a color value from the config.
+     *
+     * @param key The config key.
```
BrianBatchelder: * @param defaultColor Default value if the key does not exist.

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+     * @return The parsed color, or defaultColor.
```
BrianBatchelder: `@return The value of the config, or default value.` would match the previous getConfig* methods.

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+     */
+    private boolean hasConfig(String key) {
+        return getConfigValue(key) != null;
+    }
+
+    /**
+     * Gets a config value.
+     *
+     * @param key The key.
+     * @return The config value.
```
BrianBatchelder: What does it return if the key doesn't exist?

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+        /**
+         * Sets the production app key and secret.
+         *
+         * @param appKey    The app key.
+         * @param appSecret The app secret.
+         * @return The config editor.
+         */
+        public ConfigEditor setProductionConfig(@NonNull String appKey, @NonNull String appSecret) {
+            editor.putString(PRODUCTION_KEY, appKey)
+                    .putString(PRODUCTION_SECRET, appSecret);
```
BrianBatchelder: weird indenting

src/android/UAirshipPlugin.java
```diff
@@ -97,46 +66,76 @@
...
-            "isAppNotificationsEnabled", "dismissMessageCenter", "dismissInboxMessage", "dismissOverlayInboxMessage");
+            "isAppNotificationsEnabled", "dismissMessageCenter", "dismissInboxMessage", "dismissOverlayInboxMessage", "setAutoLaunchDefaultMessageCenter");
+
+
+    /**
+     * List of Cordova "actions". To extend the plugin, add the action below and then define the method
+     * with the signature `void <CORDOVA_ACTION>(JSONArray data, final CallbackContext callbackContext)`
+     * and it will automatically be called. All methods will be executed in the ExecutorService. Any
+     * exceptions thrown by the actions are automatically caught and the callbackContext will return
+     * an error result.
```
BrianBatchelder: Lines 82-86 seem like an unnecessary copy of lines 64-68.

src/android/PluginManager.java
```diff
@@ -0,0 +1,556 @@
...
+        /**
+         * Sets the development app key and secret.
+         *
+         * @param appKey    The app key.
+         * @param appSecret The app secret.
+         * @return The config editor.
+         */
+        public ConfigEditor setDevelopmentConfig(@NonNull String appKey, @NonNull String appSecret) {
+            editor.putString(DEVELOPMENT_KEY, appKey)
+                    .putString(DEVELOPMENT_SECRET, appSecret);
```
BrianBatchelder: weird indenting

src/android/UAirshipPlugin.java
```diff
@@ -195,14 +194,80 @@ public void onEvent(Event event) {
...
+                .setNotificationIcon(config.optString("icon"))
+                .setNotificationLargeIcon(config.optString("largeIcon"))
+                .setNotificationAccentColor(config.optString("accentColor"))
+                .apply();
+
+        callbackContext.success();
+    }
+
+    /**
+     * Enables/disables auto launching the message center.
```
BrianBatchelder: center _from a push_.

src/android/UAirshipPlugin.java
```diff
@@ -489,13 +554,12 @@ void getTags(JSONArray data, CallbackContext callbackContext) throws JSONExcepti
     /**
      * Returns the alias.
      *
+     * @param data            The call data.
+     * @param callbackContext The callback context.
      * @deprecated Deprecated since 6.7.0 - to be removed in a future version of the plugin - please use getNamedUser
-     *
+     * <p>
```
BrianBatchelder: seems weird to add `<p>` and not `</p>`, especially with `<p/>` as the next line.

src/android/UAirshipPlugin.java
```diff
@@ -755,10 +818,10 @@ void editChannelTagGroups(JSONArray data, CallbackContext callbackContext) throw
 
     /**
      * Runs an Urban Airship action.
-     *
+     * <p>
```
BrianBatchelder: seems weird to add `<p>` and not `</p>`

src/android/UAirshipPlugin.java
rlepinski: Not specifying from a push anymore

src/android/PluginManager.java
rlepinski: Added

src/android/ConfigUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String UA_PREFIX = "com.urbanairship";
+    private static final String SENDER_PREFIX = "sender:";
+
+    /**
+     * Convert the log level string to an int.
+     *
+     * @param logLevel        The log level as a string.
```
marc-scig: super minor whitespace

rlepinski: This is actually the default style in AS

PR #242, opened 2018-07-24
MOBILE-238 – Support authorized notification settings

marc-scig:
### What do these changes do?
Adds support for authorized notification settings in  the cordova plugin (iOS only) and updates iOS SDK to 9.3.2, so that they are actually available. 

Got the plugin and sample app building and running in the Simulator. Still need to test it out on a real device to make sure it's working as intended, but it's going OK so far.

Additional comments in the diff.


src/ios/UACordovaPluginManager.m
```diff
@@ -276,32 +283,56 @@ - (void)registrationFailed {
...
+
+    optedIn = authorizedSettings != UAAuthorizedNotificationSettingsNone;
 
     NSDictionary *eventBody = @{  @"optIn": @(optedIn),
+                                  // Deprecated payload for backwards compatibility
                                   @"notificationOptions" : @{
                                           NotificationPresentationAlertKey : @(alertBool),
                                           NotificationPresentationBadgeKey : @(badgeBool),
-                                          NotificationPresentationSoundKey : @(soundBool) }
+                                          NotificationPresentationSoundKey : @(soundBool)
```
marc-scig: I think that the way we've had this set up for a while a bug, or at least it's in conflict with the API doc in UrbanAirship.js. These are notification presentation keys that are intended for use in the config, unless I'm mistaken. Leaving it as it is, since at this point it's way too late to fix.

www/UrbanAirship.js
```diff
@@ -172,10 +172,13 @@ module.exports = {
    * @event UrbanAirship#notification_opt_in_status
    * @type {object}
    * @param {boolean} optIn If the user is opted in or not to user notifications.
-   * @param {object} [notificationOptions] iOS only. A map of opted in options.
```
marc-scig: I thought about keeping this and adding a deprecation note, but as far as I can tell, given the above, this never worked the way it's documented anyway. Thoughts?

src/ios/UACordovaPluginManager.m
BrianBatchelder: It would be nice if we could just pass the enum through, so we didn't have to update this every time it changed.

marc-scig: I'm open to other ideas, I was just following the pattern in place. My guess is that the reason why we never passed an enum to begin with is that JS doesn't really have them, at least not in the sense we're used to. We'd have to pass raw integers through the bridge and reconstitute then at the JS layer in some way, and I'm not sure how idiomatic option bitmasks are in that world.

rlepinski: We could keep with the current pattern or pass back an array of settings: `["alert", "badge", "blah"]`

BrianBatchelder:
Probably just a new version of Xcode.

BrianBatchelder:
Sorry about closing and re-opening. My finger slipped in the worst possible way!

crow:
🍔

PR #243, opened 2018-07-26
[MOBILE-237] Cordova: Add setter for foreground presentation options

crow:
### What do these changes do?
This adds a setter for presentation options that's accessible from the JS layer/

These changes are necessary for users to dynamically set the presentation options from the JS layer and eventually move away from setting presentation options in the config entirely. 

Added the new presentation options setter call to the quietTimeEnabled button (temporarily) and traced execution through the expected plugin and plugin manager calls.

screen shot 2018-07-25 at 4 59 46 pm
screen shot 2018-07-25 at 5 12 26 pm


rlepinski:
Do we wan to do the odd enum map thing?

rlepinski:
👍

PR #253, opened 2018-09-12
Release 7.2.1

crow:
### What do these changes do?
Update plugin to 7.2.1

To update iOS and Android SDKs with latest changes - including some important bug fixed.

Ran the samples, manually tested android to confirm [fix](https://github.com/urbanairship/urbanairship-cordova/pull/254)

Had to use git lfs to track the static lib because its size just passed the 100mb limit.


BrianBatchelder:
Can you explain more about what you had to do because the static lib got too big?

PR #254, opened 2018-09-17
Fixing issue on init firebese in Android

psabater:
### What do these changes do?

Refactor 'ext.postBuildExtras' to use 'cdvPluginPostBuildExtras.push'

When you build a cordova project with this urban plugin, GoogleServicesPlugin is not executed on android build.

This cause that firebase properties, from 'google-services.json', are not parsed into 'strings.xml'. And when you launch the application, urban fails on register into Firebase.

Run an android project without the fix. On launch android app we will see in the logcat next message:
`UALib: Unable to register with FCM: Defatult FirebaseApp is not initialized in the process <package name>. Make sure to call FirebaseApp.initializedApp(Context) first.`

This is showed by the 'autopilot' because it can't find firebase properties and firebase register fails.

Next, run the app with the fix. You will see on build gradle that the function `:app:processDebugGoogleServices` is executed and the error is disappeared in the logcat.

Before fix:
build before fix
logcat before fix

After fix:
build after fix

It seems cordova use this method to apply plugins. It will be execute in this line from `build.gradle` app:
```
for (def func : cdvPluginPostBuildExtras) {
    func()
}
```
The workaround we are currently using is to apply this plugin in our project's root `build-extras.gradle`:
```
// This can be defined within build-extras.gradle as:
//     ext.postBuildExtras = { ... code here ... }
if (hasProperty('postBuildExtras')) {
    postBuildExtras()
}
```
But is somewhat a hack and we suppose it would be better to fix it in the plugin.

### DIFF
--- a/src/android/build-extras.gradle
+++ b/src/android/build-extras.gradle
@@ -27,6 +27,6 @@ dependencies {
     implementation 'com.urbanairship.android:urbanairship-fcm:9.4.1'
 }
 
-ext.postBuildExtras = {
+cdvPluginPostBuildExtras.push({
     apply plugin: com.google.gms.googleservices.GoogleServicesPlugin
-}
+})


src/android/build-extras.gradle
```diff
@@ -27,6 +27,6 @@ dependencies {
     implementation 'com.urbanairship.android:urbanairship-fcm:9.4.1'
 }
 
-ext.postBuildExtras = {
+cdvPluginPostBuildExtras.push({
```
rlepinski: Should we namespace this to avoid conflicts with other plugins? Maybe `cdvPluginPostBuildExtras.urbanAirship`

rlepinski:
@psabater Do you know what version Cordova added support for `cdvPluginPostBuildExtras` ?

src/android/build-extras.gradle
psabater: `cdvPluginPostBuildExtras` is an array of functions. 

```
...
// Plugin gradle extensions can append to this to have code run at the end.
cdvPluginPostBuildExtras = []
...

for (def func : cdvPluginPostBuildExtras) {
    func()
}
```
If you want we can set a method name with something like it:

```
ext.urbanAirship = {
    apply plugin: com.google.gms.googleservices.GoogleServicesPlugin
}

cdvPluginPostBuildExtras.push(urbanAirship)
```

rlepinski:
We should be able to get this in today. Thanks for the PR!

Have you filled out a [contribution agreement yet?](https://docs.google.com/forms/d/e/1FAIpQLScErfiz-fXSPpVZ9r8Di2Tr2xDFxt5MgzUel0__9vqUgvko7Q/viewform)

psabater:
> We should be able to get this in today. Thanks for the PR!
> 
> Have you filled out a [contribution agreement yet?](https://docs.google.com/forms/d/e/1FAIpQLScErfiz-fXSPpVZ9r8Di2Tr2xDFxt5MgzUel0__9vqUgvko7Q/viewform)

Sorry for delay. Yes i did when you put the link.

Regards.

PR #257, opened 2018-10-03
Specify inferred types for backwards compatibility, fix reference to Context

chrisweight:
# Please fill out this template when submitting a pull request.
All lines beginning with an ℹ symbol indicate information that's important for you to provide to ensure your pull request is reviewed as quickly and efficiently as possible.

### What do these changes do?
Adds explicit types to any `<>` declarations to allow older Cordova Android (<= 6.4.0) to build
Changes `cordova.getContext()` to `cordova.getActivity().getApplicationContext()`.

To allow builds against `cordova-android@<=6.4.0`

Forked the repo, made the changes, built the app.

### DIFF
--- a/src/android/UAirshipPlugin.java
+++ b/src/android/UAirshipPlugin.java
@@ -87,7 +87,7 @@ public class UAirshipPlugin extends CordovaPlugin {
         super.initialize(cordova, webView);
         Logger.info("Initializing Urban Airship cordova plugin.");
 
-        context = cordova.getContext().getApplicationContext();
+        context = cordova.getActivity().getApplicationContext();
         pluginManager = PluginManager.shared(context);
     }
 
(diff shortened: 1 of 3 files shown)


src/android/UAirshipPlugin.java
```diff
@@ -87,7 +87,7 @@ public void initialize(CordovaInterface cordova, CordovaWebView webView) {
         super.initialize(cordova, webView);
         Logger.info("Initializing Urban Airship cordova plugin.");
 
-        context = cordova.getContext().getApplicationContext();
+        context = cordova.getActivity().getApplicationContext();
```
rlepinski: This is the only change i am worried about, we need to make sure its not deprecated in the latest release. If it is we have other ways of accessing a context.

chrisweight: Hmm, just checked the cordova-android repo source and there's no impending deprecation warning in the comments...

rlepinski: Thanks for checking!

rlepinski:
👍 Looks good to me

rlepinski:
Ill try to get a release out today. Thanks!

rlepinski:
Was not able to get to the release today. Ill try for Monday.

PR #272, opened 2018-11-30
[MOBILE-359] add update_cordova script

crow:
### What do these changes do?
Adds an update script that updates various version and also handles iOS header updates

To improve cordova update process and remove the additional step of properly merging the iOS header files.

Ran the script, observed changes. See attached video.

https://slack-files.com/T025Q1VP7-FEFNNJF0T-8f5ca130c9

### DIFF
--- a/scripts/create_sample.sh
+++ b/scripts/create_sample.sh
@@ -23,6 +23,9 @@ fi
 # keep cordova up to date
 #npm install cordova -g
 
+# Update ios static lib
+source $ROOT_PATH/scripts/update_ios_static_lib.sh
+
 # create cordova directory
 mkdir -p $CORDOVA_PATH
 cd $CORDOVA_PATH
--- a/scripts/update_cordova/index.js
+++ b/scripts/update_cordova/index.js
@@ -1,0 +1,190 @@
+#!/usr/bin/env node
+
+var co = require('co'),
+    prompt = require('co-prompt'),
+    program = require('commander'),
+    fs = require('fs'),
+    admZip = require('adm-zip'),
+    request = require('superagent'),
+    semver = require('semver');
+
+const pluginXMLVersionRegex = /version=\"[0-9]+\.[0-9]+\.[0-9]+\"/g;
+const pluginJSONVersionRegex = /"version": ?"[0-9]+\.[0-9]+\.[0-9]+"/g;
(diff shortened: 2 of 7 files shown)


scripts/create_sample.sh
```diff
@@ -23,6 +23,9 @@ fi
 # keep cordova up to date
 #npm install cordova -g
 
+# Update ios static lib
+source $ROOT_PATH/scripts/update_ios_static_lib.sh
```
BrianBatchelder: Why weren't we doing this already? Is there any problem with doing it each time we run `create_sample.sh`?

scripts/update_cordova/index.js
```diff
@@ -0,0 +1,190 @@
...
+const pluginXMLFilePath = `${__dirname}/../../plugin.xml`
+const packageJSONFilePath = `${__dirname}/../../package.json`
+const buildExtrasFilePath = `${__dirname}/../../src/android/build-extras.gradle`
+
+ program
+    .arguments('<Plugin version>')
+    .option('-a, --android <Android version>', 'The UA Android SDK version to update to.')
+    .option('-i, --ios <iOS version>', 'The UA iOS SDK version to update to.')
+    .action(function(pluginVersion) {
+            // This section will only be entered if options are included
```
BrianBatchelder: If there are arguments, but no options, does it run? It looks like it needs to run in that case (lines 28-32)

scripts/create_sample.sh
crow: No problem with running it each time, it's just helpful if you frequently clean your workspace since the static library isn't tracked

scripts/update_cordova/index.js
crow: I meant arguments, not options, fixed

scripts/update_cordova/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "update_cordova",
+  "version": "1.0.0",
+  "description": "A script for updating the cordova plugin",
+  "main": "index.js",
+  "bin": {
+    "update_cordova": "./index.js"
+  },
+  "dependencies": {
```
rlepinski: you can move this into the main package as devDependencies

crow: updated

PR #273, opened 2018-11-30
MOBILE-391: Document + Fix JS takeOff

rlepinski:
### What do these changes do?
Updates the readme with JS takeOff and configure steps. Also fixed a takeOff error log due to calling takeOff on a bg thread on Android.

People want to control everything through JS instead of xml

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,68 +29,14 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
 
         cordova plugin add urbanairship-cordova
 
-
-2. Modify the config.xml file to contain (replacing with your configuration settings):
-
-        <!-- Urban Airship app credentials -->
-        <preference name="com.urbanairship.production_app_key" value="Your Production App Key" />
-        <preference name="com.urbanairship.production_app_secret" value="Your Production App Secret" />
-        <preference name="com.urbanairship.development_app_key" value="Your Development App Key" />
-        <preference name="com.urbanairship.development_app_secret" value="Your Development App Secret" />
-
-        <!-- If the app is in production or not -->
-        <preference name="com.urbanairship.in_production" value="true | false" />
-
-        <!-- Optional config values -->
-
-        <!-- Enable push when the application launches -->
-        <preference name="com.urbanairship.enable_push_onlaunch" value="true | false" />
-
-        <!-- Enable Analytics when the application launches -->
-        <!-- Warning: Features that depend on analytics being enabled may not work properly if analytics is disabled (reports, location segmentation, region triggers, push to local time). -->
-        <preference name="com.urbanairship.enable_analytics" value="true | false" />
-
-        <!-- Urban Airship development log level defaults to debug -->
-        <preference name="com.urbanairship.development_log_level" value="none | error | warn | info | debug | verbose" />
-
-        <!-- Urban Airship production log level defaults to error -->
-        <preference name="com.urbanairship.production_log_level" value="none | error | warn | info | debug | verbose" />
-
-        <!-- Override the Android notification icon -->
-        <preference name="com.urbanairship.notification_icon" value="ic_notification" />
-
-        <!-- Override the Android notification large icon -->
-        <preference name="com.urbanairship.notification_large_icon" value="ic_notification_large" />
-
-        <!-- Override the Android notification sound (sound file should be in res/raw)-->
-        <preference name="com.urbanairship.notification_sound" value="push" />
-
-        <!-- Specify the notification accent color for Android API 21+ (Lollipop) -->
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
 
         cordova plugin add urbanairship-cordova
 
+2. *(Android Only)* Add a reference to your google-servies.json file (parallel to config.xml):
```
BrianBatchelder: Does `parallel` mean `in the same directory as`? If so, maybe that would be clearer?

README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
...
+                    "developmentTeam": "XXXXXXXXXX"
+                },
+                "release": {
+                    "developmentTeam": "XXXXXXXXXX"
+                }
+            }
+        }
+    Your iOS builds will need to reference the build.json using Cordova's "--buildConfig" flag.
+
+4. Initailize Urban Airship
```
BrianBatchelder: _Initialize_

README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
...
+          accentColor: "#FF0000"
+        })
+
+        // Configure iOS
+        UAirship.setPresentationOptions(
+          UAirship.presentationOptions.sound | UAirship.presentationOptions.alert
+        )
+
+
+    Alternatively you can configure Urban Airship through config.xml and it will takeOff automatically.
```
BrianBatchelder: Is JS now our recommended takeoff mechanism? If not, this line should probably be before the JS lines.

README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
...
-
         <!-- Urban Airship development log level defaults to debug -->
         <preference name="com.urbanairship.development_log_level" value="none | error | warn | info | debug | verbose" />
 
         <!-- Urban Airship production log level defaults to error -->
         <preference name="com.urbanairship.production_log_level" value="none | error | warn | info | debug | verbose" />
 
-        <!-- Override the Android notification icon -->
-        <preference name="com.urbanairship.notification_icon" value="ic_notification" />
+        <!-- Enables/disables auto launching the message center when the corresponding push is opened. -->
```
BrianBatchelder: Should you indicate what happens if the value is not set, as you did for the `in_production` flag?

README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
...
-                "debug": {
-                    "developmentTeam": "XXXXXXXXXX"
-                },
-                "release": {
-                    "developmentTeam": "XXXXXXXXXX"
-                }
-            }
-        }
-    Your iOS builds will need to reference the build.json using Cordova's "--buildConfig" flag.
+    `UrbanAirship.takeOff` can be called multiple times but any changes the app credentials will not apply until the next app start.
```
BrianBatchelder: any changes _to_ the app credentials

README.md
```diff
@@ -29,88 +29,112 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
...
-4. Enable user notifications:
+5. Enable user notifications:
 
-        // Enable user notifications (will prompt the user to accept push notifications)
+        // Enable user notifications (will prompt the user to accept push notifications on iOS)
         UAirship.setUserNotificationsEnabled(true, function (enabled) {
             console.log("User notifications are enabled! Fire away!")
         })
 
+6. Listen for events:
```
BrianBatchelder: Maybe indicate if this step is optional or required?

README.md
rlepinski: Going to wait on describing this further. We need to add a feature from react-native before this is really that useful. Right now cordova users need to parse the push message for the message ID. Its going to be differrent per platform. On React we send a special even when its disabled - https://docs.urbanairship.com/reference/libraries/react-native/latest/UrbanAirship.html#event:showInbox

README.md
```diff
@@ -29,88 +29,111 @@ Please visit http://support.urbanairship.com/ for any issues integrating or usin
 
         cordova plugin add urbanairship-cordova
 
+2. *(Android Only)* Add a reference to your google-servies.json file in the app's `config.xml`:
```
BrianBatchelder: _google-services.json_

README.md
BrianBatchelder: OK

README.md
BrianBatchelder: OK

crow:
🍔 sure

PR #275, opened 2018-12-14
[MOBILE-406] active notification API

marc-scig:
This adds the ability to get active (posted) notifications to our cordova plugin, as well as clear notifications, either wholesale or by identifier. The overall structure of the solution is modeled after our react module, which I borrowed a bit of code from where appropriate.

The way we've been doing this in react has been to pass up notification identifiers as strings on Android, and when appropriate affixing the notification tag if it has one. The resulting strings look like "<ID>:<TAG>", and have to be re-parsed into their constituent parts when calling getActiveNotifications. This has the advantage of making a relatively uniform API, since iOS notifications have UUIDs for identifiers. In order to do this I deprecated the notification_id field in the push event and added a notificationId field, which will always be a string regardless of platform. This matches the react API.

Tested manually in all the janky ways one would expect for a cordova project. Things seems to be humming along nicely.

### DIFF
--- a/src/android/UAirshipPlugin.java
+++ b/src/android/UAirshipPlugin.java
@@ -11,8 +11,11 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
+import android.service.notification.StatusBarNotification;
+import android.support.v4.app.NotificationManagerCompat;
 import android.support.v4.content.ContextCompat;
 
+import com.urbanairship.Autopilot;
 import com.urbanairship.Logger;
 import com.urbanairship.UAirship;
 import com.urbanairship.actions.ActionArguments;
@@ -23,6 +26,7 @@ import com.urbanairship.actions.ActionValueException;
 import com.urbanairship.cordova.events.DeepLinkEvent;
 import com.urbanairship.cordova.events.Event;
 import com.urbanairship.cordova.events.NotificationOpenedEvent;
+import com.urbanairship.cordova.events.PushEvent;
 import com.urbanairship.google.PlayServicesUtils;
 import com.urbanairship.json.JsonValue;
 import com.urbanairship.push.PushMessage;
@@ -31,7 +35,6 @@ import com.urbanairship.richpush.RichPushInbox;
 import com.urbanairship.richpush.RichPushMessage;
 import com.urbanairship.util.HelperActivity;
 import com.urbanairship.util.UAStringUtil;
-import com.urbanairship.Autopilot;
 
 import org.apache.cordova.CallbackContext;
 import org.apache.cordova.CordovaInterface;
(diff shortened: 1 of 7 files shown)


src/ios/UAirshipPlugin.m
```diff
@@ -718,6 +718,49 @@ - (void)refreshInbox:(CDVInvokedUrlCommand *)command {
     }];
 }
 
+- (void)getActiveNotifications:(CDVInvokedUrlCommand *)command {
+    [self performCallbackWithCommand:command withBlock:^(NSArray *args, UACordovaCompletionHandler completionHandler) {
+        if (@available(iOS 10.0, *)) {
```
crow: Do we need to check for 10 anymore? Isn't it the lowest supported iOS version?

marc-scig: oh good question. that was a copy pasta, probably can be removed

crow:
🍔

src/android/UAirshipPlugin.java
```diff
@@ -1195,4 +1199,83 @@ void isAppNotificationsEnabled(JSONArray data, final CallbackContext callbackCon
...
+            NotificationManager notificationManager = (NotificationManager) UAirship.getApplicationContext().getSystemService(Context.NOTIFICATION_SERVICE);
+            StatusBarNotification[] statusBarNotifications = notificationManager.getActiveNotifications();
+
+            for (StatusBarNotification statusBarNotification : statusBarNotifications) {
+                int id = statusBarNotification.getId();
+                String tag = statusBarNotification.getTag();
+
+                PushMessage pushMessage;
+                Bundle extras = statusBarNotification.getNotification().extras;
+                if (extras != null && extras.containsKey("push_message")) {
```
rlepinski: This part is non standard. Its only available in react because we override the default notification factory and stick it in there. The same can be achieved in cordova:

```
 DefaultNotificationFactory notificationFactory = new DefaultNotificationFactory(UAirship.getApplicationContext()) {
            @Override
            public NotificationCompat.Builder extendBuilder(@NonNull NotificationCompat.Builder builder, @NonNull PushMessage message, int notificationId) {
                builder.getExtras().putBundle("push_message", message.getPushBundle());
                return builder;
            }
        };
```

Put that wherever we handle the take Off onReady callback

www/UrbanAirship.js
```diff
@@ -161,7 +162,8 @@ module.exports = {
    * @type {object}
    * @param {string} message The push alert message.
    * @param {object} extras Any push extras.
-   * @param {number} [notification_id] The Android notification ID.
+   * @param {number} [notification_id] The Android notification ID. Deprecated in favor of notificationId.
+   * @param {string} [notificationId] The notification ID.
    * @param {string} [actionID] The ID of the notification action button if available.
```
rlepinski: we are so inconsistent with our naming. Should this be `actionId` or should we do `notificationID` ?

www/UrbanAirship.js
```diff
@@ -739,6 +741,41 @@ module.exports = {
...
+   * @param {function(message)} [failure] Failure callback.
+   * @param {string} failure.message The error message.
+   */
+  clearNotifications: function(success, failure) {
+    argscheck.checkArgs('FF', 'UAirship.clearNotifications', arguments)
+    callNative(success, failure, "clearNotifications")
+  },
+
+  /**
+   * Gets currently active notifications.
```
rlepinski: probably mention the android os version support

rlepinski:
👍

src/android/UAirshipPlugin.java
marc-scig: extended the CordovaNotificationFactory

PR #281, opened 2018-12-28
fix lib update script to be executable from any directory

crow:
### What do these changes do?
Fix directory handling in the ios sdk updating script

So we can execute the script from any directory without changes in behavior

Ran the script.


rlepinski:
👍

marc-scig:
👍 after fixing merge conflicts

PR #282, opened 2018-12-28
[MOBILE-412] Add showInbox event to iOS and Android implementations

crow:
### What do these changes do?
This adds a showInbox event to both iOS and Android that gets emitted if the default auto displaying inbox is disabled. 

To achieve functional parity with other similar plugins.

Ran the sample.

### DIFF
--- a/src/android/PluginManager.java
+++ b/src/android/PluginManager.java
@@ -22,6 +22,7 @@ import com.urbanairship.cordova.events.NotificationOpenedEvent;
 import com.urbanairship.cordova.events.NotificationOptInEvent;
 import com.urbanairship.cordova.events.PushEvent;
 import com.urbanairship.cordova.events.RegistrationEvent;
+import com.urbanairship.cordova.events.ShowInboxEvent;
 import com.urbanairship.push.PushMessage;
 import com.urbanairship.util.UAStringUtil;
 
@@ -66,6 +67,7 @@ public class PluginManager {
 
     private NotificationOpenedEvent notificationOpenedEvent;
     private DeepLinkEvent deepLinkEvent = null;
+    private ShowInboxEvent showInboxEvent = null;
     private Listener listener = null;
     private List<Event> pendingEvents = new ArrayList<Event>();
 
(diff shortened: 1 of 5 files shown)


src/android/PluginManager.java
```diff
@@ -128,6 +130,20 @@ public void deepLinkReceived(String deepLink) {
...
     }
 
+    /**
+     * Called to open the inbox when auto launch is disabled.
+     *
+     * @param showInboxEvent The show inbox event.
+     */
+    public void sendShowInboxEvent(ShowInboxEvent showInboxEvent) {
+        synchronized (lock) {
+            this.showInboxEvent = showInboxEvent;
```
rlepinski: why do you need to cache the event?

crow: will rm

src/ios/UACordovaPluginManager.m
```diff
@@ -193,6 +194,8 @@ - (void)showMessageForID:(NSString *)messageID {
 - (void)showInbox {
     if (self.autoLaunchMessageCenter) {
         [[UAirship messageCenter] display];
+    } else {
+        [self fireEvent:EventShowInbox data:@{}];
```
crow: should pass in in optional message id

crow: https://github.com/urbanairship/react-native-module/blob/master/ios/UARCTModule/UARCTMessageCenter.m#L37 
https://github.com/urbanairship/react-native-module/blob/master/ios/UARCTModule/UARCTEventEmitter.m#L213
Do we want to get the react-native plugin to accept a message ID too? It doesn't look like it currently does. This reflects what the react-native module currently does

crow: Will add to cordova and get a ticket together to add this to react

src/android/events/ShowInboxEvent.java
```diff
@@ -0,0 +1,51 @@
+/* Copyright 2017 Urban Airship and Contributors */
```
BrianBatchelder: 2019

rlepinski:
👍

PR #285, opened 2019-01-22
Address builds that don't get our plugin from npm registry

BrianBatchelder:
### What do these changes do?
Fix local `npm install` to also run our `update_ios_static_lib.sh` script so installs from our GitHub source create a complete cordova plugin.

It appears that Outsystems may be getting our plugin from our GitHub repo and installing it manually. The generated package is missing our static library. This change sets up the static library when `npm install` is run in a checkout of our repo.

Ran `npm install` locally. Updated our CI scripts to run `npm install` in the checkout of our repo.

This change is driven by several issues reported to us: [iOS build failed](https://github.com/urbanairship/urbanairship-cordova/issues/283) and [src/ios/Airship/libUAirship-10.0.3.a not found](https://github.com/urbanairship/urbanairship-cordova/issues/284).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "urbanairship-cordova",
-  "version": "7.3.4",
+  "version": "7.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -34,7 +34,7 @@
       "integrity": "sha1-mMyJDKZS3S7w5ws3klMQ/56Q/Is=",
       "dev": true,
       "requires": {
-        "underscore-contrib": "0.3.0"
+        "underscore-contrib": "~0.3.0"
       }
     },
     "co": {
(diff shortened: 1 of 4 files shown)


rlepinski:
👍

src/android/build-extras.gradle
```diff
@@ -19,14 +19,15 @@ allprojects {
 }
 
 dependencies {
-    implementation 'com.google.firebase:firebase-core:16.0.3'
```
BrianBatchelder: I'm a little concerned about dropping `firebase-core` here, but it should be picked up by `urbanairship-fcm` below, right?

marc-scig: it should be fine, we removed it in the native sample in 9.6 for the same reason: https://github.com/urbanairship/android-library-dev/commit/4746f26fb19728702778d9cc38dc681de25830d0#diff-04c6e90faac2675aa89e2176d2eec7d8L32

rlepinski:
👍

PR #290, opened 2019-02-21
[MOBILE-462] remove copyright date ranges

crow:
### What do these changes do?
remove copyright ranges

class headers can be rangeless, we don't want to edit these every year.


rlepinski:
👍

marc-scig:
🇧🇷

PR #291, opened 2019-02-26
Fix: Maintains 1.6 compatibility for older Cordova projects

chrisweight:
# Please fill out this template when submitting a pull request.
All lines beginning with an ℹ symbol indicate information that's important for you to provide to ensure your pull request is reviewed as quickly and efficiently as possible.

### What do these changes do?
Adds explicit types to diamond operators and removes string switch to allow plugin to build against older Cordova projects

1.6 does not support implied types in diamond operators or string switches

Created a fork, made the changes, built a legacy project using these changes, verified the project worked.

ℹ If applicable, please provide screenshots here.

ℹ Please provide any additional notes for the reviewer here.

### DIFF
--- a/src/android/ConfigUtils.java
+++ b/src/android/ConfigUtils.java
@@ -30,7 +30,7 @@ class ConfigUtils {
         if (logLevel == null || logLevel.length() == 0) {
             return defaultLogLevel;
         }
-				String logString = logLevel.trim().toLowerCase();
+        String logString = logLevel.trim().toLowerCase();
 				if (logString.equals("verbose")) {
 					return Log.VERBOSE;
 				} else if (logString.equals("debug")) {
--- a/src/android/ConfigUtils.java
+++ b/src/android/ConfigUtils.java
@@ -16,93 +16,93 @@ import java.util.Map;
  * Config Utils.
  */
 class ConfigUtils {
-    private static final String UA_PREFIX = "com.urbanairship";
-    private static final String SENDER_PREFIX = "sender:";
+  private static final String UA_PREFIX = "com.urbanairship";
+  private static final String SENDER_PREFIX = "sender:";
 
-    /**
-     * Convert the log level string to an int.
-     *
-     * @param logLevel        The log level as a string.
-     * @param defaultLogLevel Default log level.
-     * @return The log level.
-     */
-    public static int parseLogLevel(@Nullable String logLevel, int defaultLogLevel) {
-        if (logLevel == null || logLevel.length() == 0) {
-            return defaultLogLevel;
-        }
-        String logString = logLevel.trim().toLowerCase();
-				if (logString.equals("verbose")) {
-					return Log.VERBOSE;
-				} else if (logString.equals("debug")) {
-					return Log.DEBUG;
-				} else if (logString.equals("info")) {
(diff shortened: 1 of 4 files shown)


rlepinski:
@chrisweight Is there no way of forcing 1.7 source compatibility ? Can you try adding ```
android {
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_7
        targetCompatibility JavaVersion.VERSION_1_7
    }
}
```

To a build.gradle file  to force it? Java 7 is 8 years old now. Ideally we should be moving to Java 8 not back to java 6.

**Comment by chrisweight** [comment]
@rlepinski - The project I'm concerned with has over 35 plugins installed and is hugely complex in all aspects. I've tried forcing 1.7 and that causes all manner of issues when building with other plugins. 

Beyond my specific concerns, I totally understand the desire to drop older platforms, SDK's etc and take advantage of more recent developments - I empathise! But I would suggest there is still a large enough consumer-base of development teams stuck with older Cordova versions (< 7) on projects that require maintenance and updates that maintaining backwards compatibility minimises impact in one place responsibly for negligible effort (none, really, given that it's a PR from an external source), rather than forcing X-number of developments teams to jump through more hoops.

**Comment by rlepinski** [comment]
@chrisweight Fair enough. Do you have any idea on when this older version of Cordova will no longer be viable? Are you going to run into any issues with the play stores new rules about targeting modern version of Android?

We will try to not break you again but we normally develop using a newer version of Cordova and the IDE is great about suggesting modernizing our java usage.

File: src/android/ConfigUtils.java [hunk 1]
```diff
@@ -31,22 +31,20 @@ public static int parseLogLevel(@Nullable String logLevel, int defaultLogLevel)
...
-            case "warn":
-                return Log.WARN;
-            case "error":
-                return Log.ERROR;
-            case "none":
-                return Log.ASSERT;
-            default:
-                return defaultLogLevel;
-        }
+				if (logString.equals("verbose")) {
```
rlepinski: Some white space issues here

rlepinski:
👍 Ill merge and release once the whitespace issue is fixed

chrisweight:
We can set the target and min SDK versions in the cordova config, no problems on that front as yet - though I don't know how long that will be true for. 

Most teams will be looking beyond 6.4 for Cordova-Android nowadays, but there are still plenty of projects out there that will be using it that can't justify the upgrade pain, I'm sure. How long is a piece of string, really?

chrisweight:
@rlepinski - I let Android Studio auto-format it and re-committed, should be sorted now?

rlepinski:
Looks like it did a little too much or maybe the spacing was already messed up. Ill merge and see if i need to fix it.

rlepinski:
Released in 7.5.2

chrisweight:
Cheers @rlepinski - FWIW, I just checked the latest Ionic CLI defaults for integrating Cordova - and even if you use Ionic 4, the `Cordova-android` dependency that is defaulted to is _still_ `@6.3.0` :(

rlepinski:
@chrisweight Thanks for the info and the PR. We will try to not break Java 1.6 compatibility in the future. Looking into making our CI catch any compat issues.

Issue #302, opened 2019-05-20
push and notification_opened events not firing on iOS 


chrisweight:
# Please fill out this template when submitting an issue.
All lines beginning with an ℹ symbol indicate information that's important for you to provide to ensure your issue is resolved as quickly and efficiently as possible.

❗For how-to inquiries involving Urban Airship functionality or use cases, please
contact (support)[https://support.urbanairship.com/hc/en-us].

# Preliminary Info

### What version of urbanairship-cordova are you using?
7.5.4

### Is your issue affecting the Android platform, iOS platform, or both?
iOS only

### What version of cordova-ios and/or cordova-android are you using?
cordova-ios@4.5.5

### What are the versions of any relevant development tools you are using?
ℹXcode 10.2.1

# Report

### What unexpected behavior are you seeing?

Here's the scenario for context: 

- I install our application on two iOS devices (iPhone XR, iPhone 6s). 
- I log in to to the application on each device, using two different user accounts (User A and User B). 
- I enable push notifications on both devices and ensure that both users are registered against UA with Channel ID's correctly identified and showing enabled. 
- Using our app, I make User A a friend of User B (similar to any social networking 'friend request' model). 
- Users are able to make 'support requests' to their friend connections. So, I create a support request from User A to User B. 
- Creating a support request generates an event on our servers, storing the data about that request for retrieval. 
- This subsequently generates a request to the UA API from our servers to send a push notification (in this case, from User A to User B) 
- Any amendments (updates, cancellation etc) to that support request would also generate a server-side push notification request to the UA API. These all follow the same structure and use consistent keys for the type of event that has occurred - whether that is support request creation, amendment or deletion. 
- When the user receives the push notification informing them of the support request being made to them (or any updates to it), they should be able to tap on the notification, this opens the application and then shows the user the support request.

We have a number of different actions that can generate push notification requests to the UA API, the support request scenario is one example. In many cases, we want to direct the user to a different tab in the application, or to show a modal containing information based upon the push notification received.

Our server logs indicate that the requests / responses to the UA API are well formed and correct, we can also see that the notifications are sent to the devices correctly and are displayed on screen as expected.

In some cases, the `urbanairship.push` and `urbanairship.notification_opened`events does not appear to trigger in the Cordova context - which means that we do not receive the data payload of the notification in the application, thus we cannot direct the user appropriately once the app has been opened via tapping the notification.

[...]

PR #303, opened 2019-05-22
[MOBILE-657] store pending events 

crow:
### What do these changes do?
These changes change how we package events to facilitate pending event storage. This matches the android implementation for the most part. Prior to this work, pending events were not being stored.

These changes are necessary to improve event reliability in situations where events would otherwise be dropped before the listener was ready to receive notifications.

This potentially addresses the following issue.
https://github.com/urbanairship/urbanairship-cordova/issues/302

Manually tested the sample.

The deep link event has a slightly different structure than the rest. This is because the way its data payload is constructed is relatively simple. In future work, we should move all the data management logic from the plugin manager to the events themselves. I've excluded said work from this pr to reduce its complexity and mitigate the extra risk.

### DIFF
--- a/src/ios/UACordovaPluginManager.h
+++ b/src/ios/UACordovaPluginManager.h
@@ -10,8 +10,13 @@
 
 /**
  * Called to notify listeners of a new or pending event.
+ *
+ * @param eventType The event type string.
+ * @param data The json payload dictionary.
+ *
+ * @return `YES` if a properly implemented listener was notified, `NO` otherwise.
  */
--(void)notifyListener:(NSString *)eventType data:(NSDictionary *)data;
+-(BOOL)notifyListener:(NSString *)eventType data:(NSDictionary *)data;
 @end
 
 /**
--- a/src/ios/UACordovaPluginManager.h
+++ b/src/ios/UACordovaPluginManager.h
@@ -14,7 +14,7 @@
  * @param eventType The event type string.
  * @param data The json payload dictionary.
  *
- * @return `YES` if a properly implemented listener was notified, `NO` otherwise.
+ * @return `YES` if a listener was notified, `NO` otherwise.
  */
 -(BOOL)notifyListener:(NSString *)eventType data:(NSDictionary *)data;
 @end
--- a/src/ios/UACordovaPluginManager.m
+++ b/src/ios/UACordovaPluginManager.m
@@ -295,6 +295,7 @@ NSString *const EventDeepLink = @"urbanairship.deep_link";
 -(void)receivedDeepLink:(NSURL *_Nonnull)url completionHandler:(void (^_Nonnull)(void))completionHandler {
     self.lastReceivedDeepLink = [url absoluteString];
     [self fireEvent:[UACordovaDeepLinkEvent eventWithDeepLink:url]];
+    completionHandler();
 }
 
 #pragma mark UARegistrationDelegate
--- a/src/ios/UACordovaPluginManager.m
+++ b/src/ios/UACordovaPluginManager.m
@@ -448,8 +448,12 @@ NSString *const EventDeepLink = @"urbanairship.deep_link";
     id strongDelegate = self.delegate;
 
     if (strongDelegate && [strongDelegate notifyListener:event.type data:event.data]) {
+        UA_LTRACE(@"Cordova plugin manager delegate notified with event of type:%@ with data:%@", event.type, event.data);
+
         return;
     }
+
+    UA_LTRACE(@"No cordova plugin manager delegate available, storing pending event of type:%@ with data:%@", event.type, event.data);
 
     // Add pending event
     [self.pendingEvents addObject:event];
@@ -459,6 +463,8 @@ NSString *const EventDeepLink = @"urbanairship.deep_link";
     _delegate = delegate;
 
     if (delegate) {
+        UA_LTRACE(@"Cordova plugin manager delegate set:%@", delegate);
+
         NSDictionary *events = [self.pendingEvents copy];
         [self.pendingEvents removeAllObjects];
 
--- a/src/ios/UACordovaPluginManager.m
+++ b/src/ios/UACordovaPluginManager.m
@@ -35,16 +35,8 @@ NSString *const AuthorizedNotificationSettingsLockScreenKey = @"lockScreen";
 NSString *const AuthorizedNotificationSettingsNotificationCenterKey = @"notificationCenter";
 
 // Events
-NSString *const EventPushReceived = @"urbanairship.push";
-NSString *const EventNotificationOpened = @"urbanairship.notification_opened";
-NSString *const EventNotificationOptInStatus = @"urbanairship.notification_opt_in_status";
-
 NSString *const CategoriesPlistPath = @"UACustomNotificationCategories";
 
-NSString *const EventShowInbox = @"urbanairship.show_inbox";
-NSString *const EventInboxUpdated = @"urbanairship.inbox_updated";
-NSString *const EventRegistration = @"urbanairship.registration";
-NSString *const EventDeepLink = @"urbanairship.deep_link";
 
 @interface UACordovaPluginManager() <UARegistrationDelegate, UAPushNotificationDelegate, UAInboxDelegate, UADeepLinkDelegate>
 @property (nonatomic, strong) NSDictionary *defaultConfig;
--- a/src/ios/UACordovaPluginManager.m
+++ b/src/ios/UACordovaPluginManager.m
@@ -456,13 +456,15 @@ NSString *const CategoriesPlistPath = @"UACustomNotificationCategories";
     _delegate = delegate;
 
     if (delegate) {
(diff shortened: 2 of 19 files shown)


chrisweight:
@crow - Ok, have tested this (noting it's marked as `WIP`) and I'm seeing the same results as in #302.

Perhaps interestingly, one this I _have_ seen is that occasionally the `push` event IS fired, however the `notification_opened` event _never_ seems to fire...

src/ios/UACordovaPluginManager.m
```diff
@@ -418,12 +421,15 @@ - (UANotificationAction *)notificationActionForCategory:(NSString *)category act
     return notificationAction;
 }
 
-
-- (void)fireEvent:(NSString *)type data:(NSDictionary *)data {
+- (void)fireEvent:(NSObject<UACordovaEvent> *)event {
```
rlepinski: Can you add trace logs here for fireEvent

src/ios/UACordovaPluginManager.m
```diff
@@ -418,12 +421,15 @@ - (UANotificationAction *)notificationActionForCategory:(NSString *)category act
...
     id strongDelegate = self.delegate;
-    if (strongDelegate) {
-        [strongDelegate notifyListener:type data:data];
+
+    if (strongDelegate && [strongDelegate notifyListener:event.type data:event.data]) {
+        return;
```
rlepinski: Add a log here

src/ios/UACordovaPluginManager.m
```diff
@@ -433,8 +439,8 @@ - (void)setDelegate:(id<UACordovaPluginManagerDelegate>)delegate {
         NSDictionary *events = [self.pendingEvents copy];
```
rlepinski: Log when the delegate is set

crow:
@chrisweight thanks for the input. I'll get this cleaned up with some more comprehensive logging and do more troubleshooting on your issue today. I appreciate your help and patience.

crow:
@chrisweight just to give you an update, I'm pretty sure I've found the issue - but it it looks like it'll need to be addressed in another PR. I'll keep you posted with our solution as it progresses. Thanks again for the help testing this.

crow:
@chrisweight The push opened event was never firing because the sdk will make the requisite notification response callback after actions have finished running. In this case, the deep link completion handler wasn't being called (see last commit) which prevented the notification response callback from ever happening. With this change I'm able to see the deep link and push opened events instead of just the deep link event.

chrisweight:
Tested this locally and it's looking good :)

Cheers @crow and @rlepinski for the attention on this

src/ios/UACordovaPluginManager.h
```diff
@@ -10,8 +10,13 @@
 
 /**
  * Called to notify listeners of a new or pending event.
+ *
+ * @param eventType The event type string.
+ * @param data The json payload dictionary.
+ *
+ * @return `YES` if a properly implemented listener was notified, `NO` otherwise.
```
BrianBatchelder: I'm not sure what is intended by "properly implemented". I'm guessing you can just leave it out and say a listener was notified.

src/ios/UACordovaPluginManager.m
```diff
@@ -418,23 +422,32 @@ - (UANotificationAction *)notificationActionForCategory:(NSString *)category act
...
+    [self.pendingEvents addObject:event];
 }
 
 - (void)setDelegate:(id<UACordovaPluginManagerDelegate>)delegate {
     _delegate = delegate;
 
     if (delegate) {
+        UA_LTRACE(@"Cordova plugin manager delegate set:%@", delegate);
+
         NSDictionary *events = [self.pendingEvents copy];
```
BrianBatchelder: Do you need to synchronize on `self.pendingEvents` here, or is everything on the same thread?

src/ios/events/UACordovaDeepLinkEvent.h
```diff
@@ -0,0 +1,37 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+extern NSString *const EventDeepLink;
+
+/**
+ * Deep link event when a new deep link is received.
+ */
+@interface UACordovaDeepLinkEvent : NSObject<UACordovaEvent>
+
```
BrianBatchelder: Because `UACordovaDeepLinkEvent` implements  `UACordovaEvent`, I think you  don't need to re-define `type` and `data` in its header file.

src/ios/events/UACordovaDeepLinkEvent.m
```diff
@@ -0,0 +1,36 @@
...
+    }
+    return self;
+}
+
++ (instancetype)eventWithDeepLink:(NSURL *)deepLink {
+    return [[UACordovaDeepLinkEvent alloc] initWithDeepLink:deepLink];
+}
+
+- (instancetype)initWithDeepLink:(NSURL *)deepLink {
+    self = [super init];
```
BrianBatchelder: I think it would be safest to call `[self init]`. Then you don't have to worry about setting the defaults in multiple places.

src/ios/events/UACordovaDeepLinkEvent.h
BrianBatchelder: Same comment for `UACordovaInboxUpdatedEvent.h`, `UACordovaNotificationOpenedEvent.h`, `UACordovaNotificationOptInEvent.h`, `UACordovaPushEvent.h`, `UACordovaRegistrationEvent.h`, `UACordovaShowInboxEvent.h`

src/ios/events/UACordovaInboxUpdatedEvent.m
```diff
@@ -0,0 +1,23 @@
+/* Copyright Urban Airship and Contributors */
+
+
```
BrianBatchelder: whitespace

src/ios/events/UACordovaInboxUpdatedEvent.m
```diff
@@ -0,0 +1,23 @@
...
+#import "UACordovaInboxUpdatedEvent.h"
+
+NSString *const EventInboxUpdated = @"urbanairship.inbox_updated";
+
+@implementation UACordovaInboxUpdatedEvent
+
++ (instancetype)event {
+    return [[UACordovaInboxUpdatedEvent alloc ] init];
```
BrianBatchelder: whitespace after alloc

src/ios/events/UACordovaDeepLinkEvent.m
BrianBatchelder: Same comment for `UACordovaNotificationOpenedEvent.m`, `UACordovaShowInboxEvent.m`.

src/ios/UACordovaPluginManager.h
crow: if the delegate callback is implemented, if it responds to selector

rlepinski: properly implemented is the wrong thing to say. If the event was sent or something

src/ios/events/UACordovaDeepLinkEvent.h
crow: I think I actually do for the setters and getters to autosynthesize

src/ios/UACordovaPluginManager.m
rlepinski: probably not a bad idea

rlepinski:
👍 after the fixes above

@chrisweight Thanks for all the help!

src/ios/events/UACordovaDeepLinkEvent.h
BrianBatchelder: Yes, but you can just synthesize them in your .m file. Take a look at `registrationDelegate` in `UAAPNSRegistration`.

src/ios/events/UACordovaDeepLinkEvent.m
```diff
@@ -0,0 +1,27 @@
...
+
+#import "UACordovaDeepLinkEvent.h"
+
+NSString *const EventDeepLink = @"urbanairship.deep_link";
+NSString *const DeepLinkKey = @"deepLink";
+
+@implementation UACordovaDeepLinkEvent
+
++ (instancetype)event {
+    return [[UACordovaDeepLinkEvent alloc] initWithDeepLink:nil];
```
BrianBatchelder: Is this maybe a bit confusing? `initWithDeepLink` is defined as non-null in the header file, but you are passing nil here. I'm kinda surprised it isn't throwing a compiler warning.

src/ios/events/UACordovaDeepLinkEvent.m
crow: removed the extra (unused) initializer

src/ios/UACordovaPluginManager.h
crow: removed it

chrisweight:
Cheers guys, much appreciated, @rlepinski  - what's the ETA on a patch release for this?

PR #304, opened 2019-05-31
[MOBILE-627] add more logging

crow:
### What do these changes do?
Adds trace logging to the primary (non-helper) method calls, listener setting and event firing and adds debug logging to takeOff and plugin initialization.
 
To make the plugin easier to debug.

Built/ran cordova sample

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -232,6 +232,7 @@ typedef void (^UACordovaExecutionBlock)(NSArray *args, UACordovaCompletionHandle
 
     [self performCallbackWithCommand:command withBlock:^(NSArray *args, UACordovaCompletionHandler completionHandler) {
         BOOL enabled = [[args objectAtIndex:0] boolValue];
+        [UAirship location].locationUpdatesEnabled = enabled;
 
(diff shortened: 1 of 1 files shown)


src/ios/UAirshipPlugin.m
```diff
@@ -211,15 +228,18 @@ - (void)setUserNotificationsEnabled:(CDVInvokedUrlCommand *)command {
 }
 
 - (void)setLocationEnabled:(CDVInvokedUrlCommand *)command {
+    UA_LTRACE("setLocationEnabled called with command arguments: %@", command.arguments);
+
     [self performCallbackWithCommand:command withBlock:^(NSArray *args, UACordovaCompletionHandler completionHandler) {
         BOOL enabled = [[args objectAtIndex:0] boolValue];
-        [UAirship location].locationUpdatesEnabled = enabled;
```
BrianBatchelder: Is this change intended?

crow: no it snuck in from me breaking these out

crow:
force push was just me removing commits I accidentally pushed to this branch instead of the parent

PR #307, opened 2019-06-10
[MOBILE-402] adds support for custom notification channels

crow:
### What do these changes do?
These changes adds support for custom notification channels. It uses the same approach as adding the notification button groups. This means the custom notification channels are loaded from XML.

To support custom notification channels.

Ran the sample with changes in place. Cross referenced with notification button groups and react-native implementation.

This won't pass tests until the following repo is public:
https://github.com/urbanairship/airship-location-cordova/tree/master/src

Extdocs documentation forthcoming.


crow:
nothing was added in the fp, just rebased on MOBILE-648 with the airship-location-cordova in the repo

PR #309, opened 2019-06-12
[MOBILE-404] Allow setting the default notification channel ID

crow:
### What do these changes do?
Allows setting the default notification channel ID in preferences both directly and from the config.

To allow setting the default notification channel ID.

generating, running the android sample

image


crow:
broke this part out, pr forthcoming

PR #316, opened 2019-09-19
[MOBILE-881] Adds EU cloud site support to Cordova

Jahdeh:
### What do these changes do?
This adds EU cloud support to Cordova.

Customers using the EU cloud site need the Cordova SDK to support it. These changes allow for that.

I edited the individual plugins for each platform in my own Cordova app and debugged to make sure that the site given (either by the config.xml file or through takeoff) is correctly parsed and set in the individual iOS and Android SDKs.

Line 325 in PluginManager.java and line 134 in UACordovaPluginManager.m are the places where the site is set in the respective standalone SDKs. I made sure that the correct values are passed to these.

To test:

Test 1:
1. Clone this branch and add it as a plugin to your Cordova project.
2. In the config.xml file, add a preference element with the name "com.urbanairship.site" and set the value to "EU".

Example: 
`<preference name="com.urbanairship.site" value="EU" />`

3. Add a breakpoint a couple of lines before the lines I mention above (Lines 325 and 134).
4. Test to make sure `setSite` on Android receives `AirshipConfigOptions.SITE_EU;` and `airshipConfig.site` is equal to the enum value `UACloudSiteEU `.

Test 2:
1. In the config.xml file, comment out the preference element (as well as any app key elements) you added in the last test.
2. Call `UAirship.takeoff()` passing an object as an argument with your app key and secret as well as key called 'site'. 

Here's an example:

```
         {
            production: {
              appKey: "<APP_KEY>",
              appSecret: "<APP_SECRET>"
            },
            development: {
              appKey: "Some app key",
              appSecret: "Some app secret"
            },
            site: "EU"
         }
```

3. Repeat steps 3 and 4 from Test 1.

### DIFF
--- a/src/android/ConfigUtils.java
+++ b/src/android/ConfigUtils.java
@@ -2,6 +2,7 @@
 
 package com.urbanairship.cordova;
 
+import com.urbanairship.AirshipConfigOptions;
 import android.content.Context;
 import android.content.res.XmlResourceParser;
 import android.support.annotation.NonNull;
--- a/src/android/ConfigUtils.java
+++ b/src/android/ConfigUtils.java
@@ -66,6 +66,21 @@ class ConfigUtils {
     return value;
   }
 
+    /**
+   * Parses a cloud site from a String.
+   *
+   * @param value The value to parse.
+   * @return The parsed site value. Defaults to US if site is null or does not match EU.
+   */
+  @NonNull
+  @AirshipConfigOptions.Site
+  public static String parseCloudSite(@Nullable String value) {
+    if (AirshipConfigOptions.SITE_EU.equalsIgnoreCase(value)) {
+      return AirshipConfigOptions.SITE_EU;
+    }
+    return AirshipConfigOptions.SITE_US;
+  }
+
   /**
    * Parses the config.xml file for any Urban Airship config.
    *
--- a/src/android/ConfigUtils.java
+++ b/src/android/ConfigUtils.java
@@ -67,7 +67,7 @@ class ConfigUtils {
     return value;
   }
 
-    /**
+  /**
    * Parses a cloud site from a String.
    *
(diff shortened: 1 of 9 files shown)


src/android/ConfigUtils.java
```diff
@@ -66,6 +67,21 @@ public static String parseSender(@Nullable String value) {
     return value;
   }
 
+    /**
```
BrianBatchelder: whitespace

src/ios/UACordovaPluginManager.m
```diff
@@ -205,6 +215,13 @@ -(NSInteger)parseLogLevel:(id)logLevel defaultLogLevel:(UALogLevel)defaultValue
     return defaultValue;
 }
 
++ (UACloudSite)parseCloudSiteString:(NSString *)site {
```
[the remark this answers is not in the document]
rlepinski: Its a util method that does not need to be attached to an instance. This is fine.

src/ios/UACordovaPluginManager.m
```diff
@@ -205,6 +215,13 @@ -(NSInteger)parseLogLevel:(id)logLevel defaultLogLevel:(UALogLevel)defaultValue
...
+    if ([CloudSiteEUString caseInsensitiveCompare:site] == NSOrderedSame) {
```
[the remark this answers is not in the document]
rlepinski: Thats internal only

src/android/ConfigUtils.java
rlepinski: You have an extra space before the `/**`

jtowle:
Thanks for contributing this, Jade!

src/android/ConfigUtils.java
Jahdeh: Fixed the whitespace

PR #319, opened 2019-10-11
[MOBILE-663] event constructor refactor

marc-scig:
We have a few events using `eventWithData:` as a convenience constructor, and this has resulted in a situation where the plugin manager needs to know a bit too much about the shape of event data for its own good. This PR refactors those classes to take only the information they need to construct their own event data internally. Aside from the constructor changes, most of the diff is just helper methods being moved around. There's also a minor change to the opt-in event format that I'll leave a comment on below.

I'm still putting this through its paces as far as manual testing goes. In the meanwhile, have a look and see if this all makes sense.

### DIFF
--- a/src/ios/events/UACordovaNotificationOptInEvent.m
+++ b/src/ios/events/UACordovaNotificationOptInEvent.m
@@ -4,19 +4,77 @@
 
 NSString *const EventNotificationOptInStatus = @"urbanairship.notification_opt_in_status";
 
+NSString *const UACordovaNotificationOptInEventAlertKey = @"alert";
+NSString *const UACordovaNotificationOptInEventBadgeKey = @"badge";
+NSString *const UACordovaNotificationOptInEventSoundKey = @"sound";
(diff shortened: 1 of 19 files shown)


src/ios/events/UACordovaNotificationOptInEvent.m
```diff
@@ -4,19 +4,77 @@
...
+    }
+
+    if (authorizedSettings & UAAuthorizedNotificationSettingsNotificationCenter) {
+        notificationCenterBool = YES;
+    }
+
+    optedIn = authorizedSettings != UAAuthorizedNotificationSettingsNone;
+
+    NSDictionary *eventBody = @{  @"optIn": @(optedIn),
+                                  @"authorizedNotificationSettings" : @{
```
marc-scig: I removed the old (deprecated) payload here because we might as well get rid of it at this point. It's undocumented, and was keying off of the namespaced config keys used by the plugin manager, which would be gross to expose here.

PR #320, opened 2019-10-11
MOBILE-896 Update cordova plugin to iOS SDK 12.0

marc-scig:
This is mostly updating the plugin and podspec versions, with a few fixes for the deprecation warnings that come with the push -> channel refactor, as well as some iOS 13 stuff. I also had to bump the minimum SDK version of the sample to 11.0 so that the SDK pod would install properly. 

While I was looking, I also cleaned up a variety of compiler and analyzer warnings such as:

* missing nullability annotations
* dictionary values that might be accidentally nil (which would cause crashes)
* the overlay inbox message method, which was accidentally recursing into itself instead of calling its helper

Manual testing is in progress but no surprises so far.

Note: because of the current CI issues, and because it would be a gnarly rebase otherwise, I'm temporarily setting this up to merge into the the event data refactor branch, which should go out with this update.

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -53,8 +53,9 @@ typedef void (^UACordovaExecutionBlock)(NSArray *args, UACordovaCompletionHandle
 
     // String
     if ([value isKindOfClass:[NSString class]]) {
+        NSCharacterSet *characters = [NSCharacterSet URLHostAllowedCharacterSet];
         return [CDVPluginResult resultWithStatus:status
-                                 messageAsString:[value stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding]];
+                                 messageAsString:[value stringByAddingPercentEncodingWithAllowedCharacters:characters]];
     }
 
     // Number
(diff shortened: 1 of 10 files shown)


src/ios/UAirshipPlugin.m
```diff
@@ -757,12 +758,46 @@ - (void)overlayInboxMessage:(CDVInvokedUrlCommand *)command {
...
+            return adapter;
+        } forDisplayType:UAInAppMessageDisplayTypeHTML];
+
+        self.factoryBlockAssigned = YES;
+    }
+
+    [UAActionRunner runActionWithName:kUAOverlayInboxMessageActionDefaultRegistryName
+                                value:message.messageID
+                            situation:UASituationManualInvocation];
+}
```
BrianBatchelder: To spark a discussion elsewhere: We are duplicating this (and other) code in all of our framework plugins. I wonder if we shouldn't look into refactoring common plugin code into some sort of library or iOS framework.

rlepinski: lets not discuss this on this PR. good topic for coworking time.

PR #323, opened 2020-01-10
[MOBILE-1042] Update to iOS 13.0.4 SDK

rlepinski:
### What do these changes do?
- Updates the iOS SDK to 13.0.4
- Removes overlay inbox message functionality since its no longer easily done in the SDK

Sample test. Manually verified message center calls still work.

### DIFF
--- a/src/ios/UAirshipPlugin.m
+++ b/src/ios/UAirshipPlugin.m
@@ -5,6 +5,7 @@
 #import "UACordovaPluginManager.h"
 #import "UACordovaPushEvent.h"
 
+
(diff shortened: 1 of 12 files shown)


rlepinski:
CI is failing because of the cordova pod issue

src/ios/UAirshipPlugin.m
```diff
@@ -5,6 +5,7 @@
 #import "UACordovaPluginManager.h"
 #import "UACordovaPushEvent.h"
 
+
```
BrianBatchelder:
```suggestion
```

PR #331, opened 2020-02-20
[MOBILE-1279] Add data collection enabled and push token registration enabled

Ulrico972:
### What do these changes do?
Add data collection enabled and push token registration enabled

To keep cordova module up to date with the Airship native library features

By adding settings in the sample and manually testing the different cases

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -192,7 +192,7 @@
                 <source url="https://github.com/CocoaPods/Specs.git"/>
             </config>
             <pods use-frameworks="true">
-                <pod name="Airship" spec="13.0.4" />
+                <pod name="Airship" spec="13.1.0" />
             </pods>
         </podspec>
 
--- a/www/UrbanAirship.js
+++ b/www/UrbanAirship.js
@@ -967,6 +967,19 @@ module.exports = {
     _runAction("add_custom_event_action", actionArg, success, failure)
   },
 
+  /**
+   * Initiates screen tracking for a specific app screen, must be called once per tracked screen.
+   *
+   * @param {string} screen The screen's string identifier.
+   * @param {function} [success] Success callback.
+   * @param {function(message)} [failure] Failure callback.
+   * @param {string} failure.message The error message.
+   */
+  trackScreen: function(screen, success, failure) {
+    argscheck.checkArgs('sFF', 'UAirship.trackScreen', arguments)
+    callNative(success, failure, "trackScreen", [screen])
+  },
+
   setDataCollectionEnabled: function(enabled, success, failure) {
     argscheck.checkArgs('*FF', 'UAirship.setDataCollectionEnabled', arguments)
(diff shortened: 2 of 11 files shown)


plugin.xml
```diff
@@ -192,7 +192,7 @@
                 <source url="https://github.com/CocoaPods/Specs.git"/>
             </config>
             <pods use-frameworks="true">
-                <pod name="Airship" spec="13.0.4" />
+                <pod name="Airship" spec="13.1.0" />
```
BrianBatchelder: You need to rebase on master. This change has already been made. There might be others.

www/UrbanAirship.js
```diff
@@ -965,5 +965,25 @@ module.exports = {
     }
 
     _runAction("add_custom_event_action", actionArg, success, failure)
+  },
+
```
BrianBatchelder: Can you add a doc block to each of these? See previous functions for examples.

plugin.xml
Ulrico972: Done

www/UrbanAirship.js
Ulrico972: Done

www/UrbanAirship.js
```diff
@@ -980,21 +980,53 @@ module.exports = {
     callNative(success, failure, "trackScreen", [screen])
   },
 
+  /**
+   * Enables or disables data collection.
+   *
+   * @param {boolean} enabled true to enable data collection, false to disable.
```
BrianBatchelder:
```suggestion
   * @param {boolean} set true to enable data collection, false to disable.
```

www/UrbanAirship.js
```diff
@@ -980,21 +980,53 @@ module.exports = {
...
+   */
   isDataCollectionEnabled: function(success, failure) {
     argscheck.checkArgs('fF', 'UAirship.isDataCollectionEnabled', arguments)
     callNative(success, failure, "isDataCollectionEnabled")
   },
 
+  /**
+   * Enables or disables push token registration.
+   *
+   * @param {boolean} enabled true to enable push token registration, false to disable.
```
BrianBatchelder:
```suggestion
   * @param {boolean} set true to enable push token registration, false to disable.
```

www/UrbanAirship.js
Ulrico972: I'm not sure we should change it, enabled is just the variable name =)

www/UrbanAirship.js
Ulrico972: Same here

BrianBatchelder: OK

www/UrbanAirship.js
BrianBatchelder: OK

BrianBatchelder:
@Ulrico972 I merged this PR so I could move on to the release.

PR #332, opened 2020-02-21
[MOBILE-1280] Add number attribute support to Cordova

BrianBatchelder:
### What do these changes do?
Add number channel attribute support to Cordova plugin for Android and iOS.

Support the attributes.

Manually set through javascript. Checked result using cool new API endpoint I got from Weasel.

image

### DIFF
--- a/src/android/UAirshipPlugin.java
+++ b/src/android/UAirshipPlugin.java
@@ -1067,10 +1067,17 @@ public class UAirshipPlugin extends CordovaPlugin {
 
             String action = operation.optString(ATTRIBUTE_OPERATION_TYPE);
             String key = operation.optString(ATTRIBUTE_OPERATION_KEY);
-            String value = operation.optString(ATTRIBUTE_OPERATION_VALUE);
 
             if (ATTRIBUTE_OPERATION_SET.equals(action)) {
-                editor.setAttribute(key, value);
+                Object value = operation.opt(ATTRIBUTE_OPERATION_VALUE);
+
+                if (value == JSONObject.NULL) {
+                    continue;
+                } else if (value instanceof String) {
+                    editor.setAttribute(key, (String) value);
+                } else if (value instanceof Number) {
+                    editor.setAttribute(key, ((Number) value).doubleValue());
+                }
             } else if (ATTRIBUTE_OPERATION_REMOVE.equals(action)) {
                 editor.removeAttribute(key);
             }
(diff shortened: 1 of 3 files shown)


www/UrbanAirship.js
```diff
@@ -125,7 +125,7 @@ function AttributesEditor(nativeMethod) {
    * @return {AttributesEditor} The attribute editor instance.
    */
   editor.setAttribute = function(name, value) {
-    argscheck.checkArgs('ss', "AttributesEditor#setAttribute", arguments)
+    argscheck.checkArgs('s*', "AttributesEditor#setAttribute", arguments)
```
marc-scig: asterisk means "anything goes", right? is that what we want?

src/android/UAirshipPlugin.java
```diff
@@ -1067,10 +1067,17 @@ void editChannelAttributes(@NonNull JSONArray data, @NonNull CallbackContext cal
 
             String action = operation.optString(ATTRIBUTE_OPERATION_TYPE);
             String key = operation.optString(ATTRIBUTE_OPERATION_KEY);
-            String value = operation.optString(ATTRIBUTE_OPERATION_VALUE);
 
             if (ATTRIBUTE_OPERATION_SET.equals(action)) {
-                editor.setAttribute(key, value);
+                Object value = operation.opt(ATTRIBUTE_OPERATION_VALUE);
+
+                if (value == JSONObject.NULL) {
```
marc-scig: Do we want to compare with actual null, here? What happens if someone passes null or undefined on the JS side to an unchecked argument?

https://developer.android.com/reference/org/json/JSONObject#opt(java.lang.String)

www/UrbanAirship.js
marc-scig: (I assume there's a good reason for it but I thought I'd ask)

BrianBatchelder: It does mean "any type". I'd prefer to indicate "number or string", but `checkArgs()` doesn't support that.

src/android/UAirshipPlugin.java
BrianBatchelder: That's how I tested this, by passing in `null`. It gets assigned to `value` as `JSONObject.NULL`. The javascript I tested was:

`UAirship.editChannelAttributes().setAttribute("teststring","testy").setAttribute("testnull",null).setAttribute("testnumber",3).apply()`

In this case we skipped setting the "testnull" attribute, which seems like the correct behavior.

marc-scig: What about undefined?

BrianBatchelder: Tested with `undefined`. It also works fine. We skip setting any attribute with value `undefined`.

PR #333, opened 2020-02-25
Mobile 1339

rlepinski:
Attempting to fix CI


scripts/create_sample.sh
```diff
@@ -20,23 +20,19 @@ fi
...
 mkdir -p $CORDOVA_PATH
 cd $CORDOVA_PATH
 
 # create the test project
 rm -rf test
-cordova create test com.urbanairship.sample Test
+npm install cordova@9.0.0
+npx cordova create test com.urbanairship.sample Test
 cd test
+npm install cordova@9.0.0
```
BrianBatchelder: Is there any way to set these version numbers in a single place? I fear they will get out of sync.

rlepinski: done

PR #336, opened 2020-02-28
[MOBILE-1342] Move deploy to github actions

rlepinski:
I copied the react deploy and made it work with cordova

### DIFF
--- a/scripts/upload_docs.sh
+++ b/scripts/upload_docs.sh
@@ -17,7 +17,7 @@ fi
 ROOT_PATH=`dirname "${0}"`/..
 TAR_NAME="$1.tar.gz"
 
-cd out
+cd $2
 tar -czf $TAR_NAME *
 cd -
 
--- a/scripts/upload_docs.sh
+++ b/scripts/upload_docs.sh
@@ -21,4 +21,4 @@ cd out
 tar -czf $TAR_NAME *
 cd -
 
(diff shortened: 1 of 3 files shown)


scripts/upload_docs.sh
```diff
@@ -0,0 +1,24 @@
...
+set -e
+set -x
+
+if [ -z "$1" ];
+then
+    echo "No version supplied"
+    exit 1
+fi
+
+if [ ! -d "$2" ];
```
BrianBatchelder: It doesn't look like you use $2 anywhere.

scripts/upload_docs.sh
```diff
@@ -0,0 +1,24 @@
...
+then
+    echo "Missing docs $2"
+    exit 1
+fi
+
+ROOT_PATH=`dirname "${0}"`/..
+TAR_NAME="$1.tar.gz"
+
+cd out
```
BrianBatchelder: I think this should be `cd "$2"`.

scripts/upload_docs.sh
```diff
@@ -0,0 +1,24 @@
...
+tar -czf $TAR_NAME *
+cd -
+
+gsutil cp $ROOT_PATH/out/$TAR_NAME gs://ua-web-ci-prod-docs-transfer/libraries/urbanairship-cordova/$TAR_NAME
```
BrianBatchelder:
```suggestion
gsutil cp $ROOT_PATH/$2/$TAR_NAME gs://ua-web-ci-prod-docs-transfer/libraries/urbanairship-cordova/$TAR_NAME
```

PR #339, opened 2020-03-16
[MOBILE-1344] Add plugin version check to CI

Ulrico972:
### What do these changes do?
Add plugin version check to cordova CI

Because we have 4 different places where we have to version our cordova plugin and it will be safer to have this check in the CI

By testing it manually with the CI script.

Capture d’écran 2020-03-16 à 12 09 07

The script can be used outside of the CI too.

### DIFF
--- a/scripts/check_plugin_version.sh
+++ b/scripts/check_plugin_version.sh
@@ -9,20 +9,32 @@ packageJSONFilePath="$ROOT_PATH/package.json"
 packageJSONVersionRegex='"version": ?"[0-9]+\.[0-9]+\.[0-9]+"'
 packageJSONVersion=$(grep -E "$packageJSONVersionRegex" $packageJSONFilePath | cut -f4 -d \")
 
-
 # Get version from plugin.xml file
 pluginXMLFilePath="$ROOT_PATH/plugin.xml"
+
 # Get plugin version
 pluginXMLFileVersionRegex='version= ?"[0-9]+\.[0-9]+\.[0-9]+"'
 pluginXMLFileVersion=$(grep -E "$pluginXMLFileVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+
+if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
+	exit 1
+fi
+
 # Get plugin version for android config
 pluginXMLAndroidConfigVersionRegex='android:value= ?"[0-9]+\.[0-9]+\.[0-9]+"'
 pluginXMLAndroidConfigVersion=$(grep -E "$pluginXMLAndroidConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+
+if [ "$pluginXMLAndroidConfigVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version in the Android Config is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
+	exit 1
+fi
+
 # Get plugin version for iOS config
 pluginXMLiOSConfigVersionRegex='<string>[0-9]+\.[0-9]+\.[0-9]+'
 pluginXMLiOSConfigVersion=$(grep -E "$pluginXMLiOSConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \> | cut -f1 -d \<)
 
-if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ] || [ "$pluginXMLAndroidConfigVersion" != "$packageJSONVersion" ] || [ "$pluginXMLiOSConfigVersion" != "$packageJSONVersion" ]; then
-	echo "BUILD FAILED: The plugin version is not the same everywhere (please check package.json and plugin.xml files)"
+if [ "$pluginXMLiOSConfigVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version in the iOS Config is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
 	exit 1
 fi
--- a/scripts/check_plugin_version.sh
+++ b/scripts/check_plugin_version.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -e
+set -x
+
+ROOT_PATH=`dirname "${0}"`/..
+
+# Get version from package.json file
+packageJSONFilePath="$ROOT_PATH/package.json"
+packageJSONVersionRegex='"version": ?"[0-9]+\.[0-9]+\.[0-9]+"'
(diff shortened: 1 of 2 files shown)


scripts/check_plugin_version.sh
```diff
@@ -0,0 +1,28 @@
...
+set -x
+
+ROOT_PATH=`dirname "${0}"`/..
+
+# Get version from package.json file
+packageJSONFilePath="$ROOT_PATH/package.json"
+packageJSONVersionRegex='"version": ?"[0-9]+\.[0-9]+\.[0-9]+"'
+packageJSONVersion=$(grep -E "$packageJSONVersionRegex" $packageJSONFilePath | cut -f4 -d \")
+
+
```
BrianBatchelder:
```suggestion
```

scripts/check_plugin_version.sh
```diff
@@ -0,0 +1,28 @@
...
+# Get version from plugin.xml file
+pluginXMLFilePath="$ROOT_PATH/plugin.xml"
+# Get plugin version
```
BrianBatchelder:
```suggestion

# Get plugin version
```

scripts/check_plugin_version.sh
```diff
@@ -0,0 +1,28 @@
...
+pluginXMLFileVersionRegex='version= ?"[0-9]+\.[0-9]+\.[0-9]+"'
+pluginXMLFileVersion=$(grep -E "$pluginXMLFileVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+# Get plugin version for android config
```
BrianBatchelder:
```suggestion

# Get plugin version for android config
```

scripts/check_plugin_version.sh
```diff
@@ -0,0 +1,28 @@
...
+pluginXMLAndroidConfigVersionRegex='android:value= ?"[0-9]+\.[0-9]+\.[0-9]+"'
+pluginXMLAndroidConfigVersion=$(grep -E "$pluginXMLAndroidConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+# Get plugin version for iOS config
```
BrianBatchelder:
```suggestion

# Get plugin version for iOS config
```

scripts/check_plugin_version.sh
```diff
@@ -0,0 +1,28 @@
...
+pluginXMLiOSConfigVersionRegex='<string>[0-9]+\.[0-9]+\.[0-9]+'
+pluginXMLiOSConfigVersion=$(grep -E "$pluginXMLiOSConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \> | cut -f1 -d \<)
+
+if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ] || [ "$pluginXMLAndroidConfigVersion" != "$packageJSONVersion" ] || [ "$pluginXMLiOSConfigVersion" != "$packageJSONVersion" ]; then
```
BrianBatchelder: I would put split these tests up and put them after you create the values, e.g. `if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ]` could go after line 17.  Then your `echo` message could be specific to which version checks are failing.

scripts/run_ci_tasks.sh
```diff
@@ -39,6 +39,10 @@ while true; do
   shift
 done
 
+# ckeck for version
+${SCRIPT_DIRECTORY}/check_plugin_version.sh
+
+
```
BrianBatchelder:
```suggestion
```

scripts/check_plugin_version.sh
Ulrico972: Ok I see. It's done. Don't hesitate if you have a better sentence for the error message.

scripts/check_plugin_version.sh
```diff
@@ -9,20 +9,32 @@ packageJSONFilePath="$ROOT_PATH/package.json"
...
-
 # Get version from plugin.xml file
 pluginXMLFilePath="$ROOT_PATH/plugin.xml"
+
 # Get plugin version
 pluginXMLFileVersionRegex='version= ?"[0-9]+\.[0-9]+\.[0-9]+"'
 pluginXMLFileVersion=$(grep -E "$pluginXMLFileVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+
+if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
```
BrianBatchelder:
```suggestion
	echo "BUILD FAILED: The plugin version is not correct in the plugin.xml file. The version should be the same as the one in the package.json file."
```

scripts/check_plugin_version.sh
```diff
@@ -9,20 +9,32 @@ packageJSONFilePath="$ROOT_PATH/package.json"
...
+	exit 1
+fi
+
 # Get plugin version for android config
 pluginXMLAndroidConfigVersionRegex='android:value= ?"[0-9]+\.[0-9]+\.[0-9]+"'
 pluginXMLAndroidConfigVersion=$(grep -E "$pluginXMLAndroidConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \")
+
+if [ "$pluginXMLAndroidConfigVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version in the Android Config is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
```
BrianBatchelder:
```suggestion
	echo "BUILD FAILED: The plugin version in the Android Config is not correct in the plugin.xml file. The version should be the same as the one in the package.json file."
```

scripts/check_plugin_version.sh
```diff
@@ -9,20 +9,32 @@ packageJSONFilePath="$ROOT_PATH/package.json"
...
+fi
+
 # Get plugin version for iOS config
 pluginXMLiOSConfigVersionRegex='<string>[0-9]+\.[0-9]+\.[0-9]+'
 pluginXMLiOSConfigVersion=$(grep -E "$pluginXMLiOSConfigVersionRegex" $pluginXMLFilePath | cut -f2 -d \> | cut -f1 -d \<)
 
-if [ "$pluginXMLFileVersion" != "$packageJSONVersion" ] || [ "$pluginXMLAndroidConfigVersion" != "$packageJSONVersion" ] || [ "$pluginXMLiOSConfigVersion" != "$packageJSONVersion" ]; then
-	echo "BUILD FAILED: The plugin version is not the same everywhere (please check package.json and plugin.xml files)"
+if [ "$pluginXMLiOSConfigVersion" != "$packageJSONVersion" ]; then
+	echo "BUILD FAILED: The plugin version in the iOS Config is not correct in the plugin.xml file. The version should be the same than the one in the package.json file."
```
BrianBatchelder:
```suggestion
	echo "BUILD FAILED: The plugin version in the iOS Config is not correct in the plugin.xml file. The version should be the same as the one in the package.json file."
```

PR #349, opened 2020-05-05
[MOBILE-1507] Add support for date attributes.

BrianBatchelder:
### What do these changes do?
Add support for date attributes to Cordova

Need full channel attribute support in all of our supported frameworks.

Set date attributes using sample apps. Checked channel API for results.

image

- The JS Date type doesn't survive through the JS to native bridge, so the code converts it to a milliseconds since the epoch and marks it as "type" "date". The native code then converts it back to its native date type and passes it to the SDK.


www/UrbanAirship.js
```diff
@@ -121,11 +121,15 @@ function AttributesEditor(nativeMethod) {
...
    *
    * @param {string} name The attribute name.
-   * @param {string} value The attribute's value.
+   * @param {string|number|Date} value The attribute's value.
    * @return {AttributesEditor} The attribute editor instance.
    */
   editor.setAttribute = function(name, value) {
     argscheck.checkArgs('s*', "AttributesEditor#setAttribute", arguments)
+    if (value instanceof Date) {
+      let valueAsISOString = value.toISOString().slice(0,19)
```
rlepinski, line 130: not a fan of this, either verify we can send a `z` in date attributes with david latour, send milliseconds and other hint information to the underlying sdk so it can parse the milliseconds as a native date, or figure out how to format the date properly using a formatter.

BrianBatchelder: OK. I'm comfortable with it because the first 19 characters of the date format we are using is fixed by the ISO standard. I'll check with David, as sending the milliseconds would be preferable.

www/UrbanAirship.js
```diff
@@ -121,14 +121,25 @@ function AttributesEditor(nativeMethod) {
...
-    return editor
+    editor.setAttribute = function(name, value) {
+        argscheck.checkArgs('s*', "AttributesEditor#setAttribute", arguments)
+        console.log("BDB: typeof value: " + typeof value)
+
+        var operation = { "action": "set", "value": value, "key": name }
+
+        // Date type doesn't survive through the JS to native bridge. Convert value to string and mark as a date.
+        if (value instanceof Date) {
+            console.log("BDB: processing date")
```
rlepinski, line 135: BDB

src/ios/UAirshipPlugin.m
```diff
@@ -819,7 +819,18 @@ - (void)editChannelAttributes:(CDVInvokedUrlCommand *)command {
 
             if ([action isEqualToString:@"set"]) {
                 id value = operation[@"value"];
-                if ([value isKindOfClass:[NSString class]]) {
+                NSString *valueType = operation[@"type"];
+                if ([valueType isEqualToString:@"date"]) {
```
rlepinski, line 823: why not use the type everywhere?

src/ios/UAirshipPlugin.m
```diff
@@ -819,7 +819,18 @@ - (void)editChannelAttributes:(CDVInvokedUrlCommand *)command {
...
+                    // Date type doesn't survive through the JS to native bridge. Value contains an ISO-8601 compliant string.
+                    NSDateFormatter* dateFormatter = [[NSDateFormatter alloc] init];
+                    dateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+                    dateFormatter.timeZone = [NSTimeZone timeZoneForSecondsFromGMT:0];
+                    dateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
```
rlepinski, line 828: Instead of this, you should send milliseconds from epoch

src/ios/UAirshipPlugin.m
BrianBatchelder: The number and string types pass through the bridge just fine. It seemed unnecessary.

src/ios/UAirshipPlugin.m
BrianBatchelder: Are you thinking that is safer?

rlepinski: Its less work to have to parse a timestamp, convert it seconds/milliseconds (date), then back to a timestamp.

src/ios/UAirshipPlugin.m
rlepinski: Its consistent in behavior and we could make assumptions about type instead of checking if its an instance of a type.

BrianBatchelder: FYI: I left the instance type checks in there as a default if the value type is missing or unknown. This will help if we missed any JS types that used to work "automagically".

src/android/UAirshipPlugin.java
```diff
@@ -1164,8 +1165,22 @@ private void editChannelAttributes(@NonNull JSONArray data, @NonNull CallbackCon
...
             if (ATTRIBUTE_OPERATION_SET.equals(action)) {
                 Object value = operation.opt(ATTRIBUTE_OPERATION_VALUE);
-
-                if (value instanceof String) {
+                String valueType = (String) operation.opt(ATTRIBUTE_OPERATION_VALUETYPE);
+                if ("date".equals(valueType)) {
+                    // Date type doesn't survive through the JS to native bridge. Value contains number of milliseconds since the epoch.
+                    try {
+                        Date date = new Date((Long) value);
+                        if (date != null) {
```
rlepinski, line 1173: I dont belive your date will ever be null

src/android/UAirshipPlugin.java
```diff
@@ -1164,8 +1165,22 @@ private void editChannelAttributes(@NonNull JSONArray data, @NonNull CallbackCon
 
             if (ATTRIBUTE_OPERATION_SET.equals(action)) {
                 Object value = operation.opt(ATTRIBUTE_OPERATION_VALUE);
-
-                if (value instanceof String) {
+                String valueType = (String) operation.opt(ATTRIBUTE_OPERATION_VALUETYPE);
+                if ("date".equals(valueType)) {
+                    // Date type doesn't survive through the JS to native bridge. Value contains number of milliseconds since the epoch.
+                    try {
+                        Date date = new Date((Long) value);
```
rlepinski, line 1172: `((Number) value).longValue()`

src/android/UAirshipPlugin.java
```diff
@@ -1164,8 +1165,22 @@ private void editChannelAttributes(@NonNull JSONArray data, @NonNull CallbackCon
...
+                        if (date != null) {
+                            editor.setAttribute(key, date);
+                        }
+                    } catch (Exception e) {
```
rlepinski, line 1176: why are you catching it here?

src/android/UAirshipPlugin.java
```diff
@@ -1164,8 +1165,22 @@ private void editChannelAttributes(@NonNull JSONArray data, @NonNull CallbackCon
...
+                        PluginLogger.warn(e, "Failed to parse date attribute: %d", value);
+                    }
+                } else if ("string".equals(valueType)) {
+                    editor.setAttribute(key, (String) value);
+                } else if ("number".equals(valueType)) {
+                    editor.setAttribute(key, ((Number) value).doubleValue());
+                } else if (value instanceof String) {
```
rlepinski, line 1183: why do we need this? should be removed since control the input

www/UrbanAirship.js
```diff
@@ -121,14 +121,33 @@ function AttributesEditor(nativeMethod) {
...
+        } else if (typeof value === "boolean") {
+            // No boolean attribute type. Convert value to string.
+            operation["type"] = "string"
+            operation["value"] = value.toString();
+        } else if (value instanceof Date) {
+            // Date type doesn't survive through the JS to native bridge. Convert value to milliseconds since epoch and mark as a date.
+            operation["type"] = "date"
+            operation["value"] = value.getTime()
+        } else {
+            operation["type"] = "unknown"
```
rlepinski, line 145: Throw an exception?

src/ios/UAirshipPlugin.m
rlepinski: I would rather throw an exception in JS or toString the value and set it as the atttribute

src/ios/UAirshipPlugin.m
```diff
@@ -819,7 +819,21 @@ - (void)editChannelAttributes:(CDVInvokedUrlCommand *)command {
 
             if ([action isEqualToString:@"set"]) {
                 id value = operation[@"value"];
-                if ([value isKindOfClass:[NSString class]]) {
+                NSString *valueType = operation[@"type"];
+                if ([valueType isEqualToString:@"date"]) {
+                    // Date type doesn't survive through the JS to native bridge. Value contains number of milliseconds since the epoch.
+                    if (![value isKindOfClass:[NSNumber class]]) {
```
rlepinski, line 825: Trust our JS is correct, no need to do this extra check

src/ios/UAirshipPlugin.m
```diff
@@ -819,7 +819,21 @@ - (void)editChannelAttributes:(CDVInvokedUrlCommand *)command {
...
+                        continue;
+                    }
+                    NSDate *date = [NSDate dateWithTimeIntervalSince1970:[(NSNumber *)value doubleValue] / 1000.0];
```
rlepinski, line 828: date will never be null

src/android/UAirshipPlugin.java
BrianBatchelder: I was trying to catch the unknown case, where a type we don't catch has been making it through the bridge and into the SDK. I could force a limited set of types in the JS, but we might break something that has already been working.

rlepinski: we are only supposed to support a known set, if we get something like class name, address, or some gibberish toString output that wont be great. We should do a better job limiting acceptable types and reducing complexity in our plugins.

src/android/UAirshipPlugin.java
BrianBatchelder: Because I only wanted to fail that set, not all of the sets.

BrianBatchelder: I guess I could have a broader catch in case any of the code that processes that particular attribute throws.

rlepinski: It gets caught in the execute. This is an edge case that you are adding too much validation against. We control the JS side of things so it should never happen.

src/ios/UAirshipPlugin.m
```diff
@@ -819,10 +819,17 @@ - (void)editChannelAttributes:(CDVInvokedUrlCommand *)command {
...
                 id value = operation[@"value"];
-                if ([value isKindOfClass:[NSString class]]) {
+                NSString *valueType = operation[@"type"];
+                if ([valueType isEqualToString:@"string"]) {
                     [mutations setString:(NSString *)value forAttribute:name];
-                } else if ([value isKindOfClass:[NSNumber class]]) {
+                } else if ([valueType isEqualToString:@"number"]) {
                     [mutations setNumber:(NSNumber *)value forAttribute:name];
+                } else if ([valueType isEqualToString:@"date"]) {
+                    // Date type doesn't survive through the JS to native bridge. Value contains number of milliseconds since the epoch.
```
marc-scig, line 828: What do you mean by "Date type doesn't survive..."? Is this just a comment on the fact that the value is in epoch ms?

www/UrbanAirship.js
```diff
@@ -121,14 +121,33 @@ function AttributesEditor(nativeMethod) {
...
+        if (typeof value === "string") {
+            operation["type"] = "string"
+        } else if (typeof value === "number") {
+            operation["type"] = "number"
+        } else if (typeof value === "boolean") {
+            // No boolean attribute type. Convert value to string.
+            operation["type"] = "string"
+            operation["value"] = value.toString();
+        } else if (value instanceof Date) {
+            // Date type doesn't survive through the JS to native bridge. Convert value to milliseconds since epoch and mark as a date.
```
marc-scig, line 141: Ah, I see. Can we just say "dates are serialized as milliseconds since epoch"?

PR #350, opened 2020-05-05
[MOBILE-1505] Update iOS SDK to 13.3.0

rlepinski:
### What do these changes do?
Updates iOS SDK to 13.3.0. The biggest change is the message view deprecations. I copied most the changes from DefaultSplitViewController. There are few things we could do to make it easier to implement without as much duplicate code in the future, like maybe an MessageViewErrorHandler.

Updates latest iOS SDK. 

Ran sample on both iOS 11 and 13 simulators.

Simulator Screen Shot - iPhone 8 - 2020-05-05 at 09 29 57

### DIFF
--- a/src/ios/UAMessageViewController.m
+++ b/src/ios/UAMessageViewController.m
@@ -2,26 +2,164 @@
 
 #import "UAMessageViewController.h"
 
+@interface UAMessageViewController() <UAMessageCenterMessageViewDelegate>
+@property (nonatomic, copy) NSString *pendingMessageID;
+@property (nonatomic, strong) UADefaultMessageCenterMessageViewController *airshipMessageViewController;
+@end
+
 @implementation UAMessageViewController
 
 - (void) viewDidLoad {
     [super viewDidLoad];
+
+    self.airshipMessageViewController = [[UADefaultMessageCenterMessageViewController alloc]
+                                         initWithNibName:@"UADefaultMessageCenterMessageViewController"
+                                         bundle:[UAMessageCenterResources bundle]];
+    self.airshipMessageViewController.delegate = self;
+
 
     UIBarButtonItem *done = [[UIBarButtonItem alloc]
                              initWithBarButtonSystemItem:UIBarButtonSystemItemDone
                              target:self
                              action:@selector(inboxMessageDone:)];
 
-    self.navigationItem.rightBarButtonItem = done;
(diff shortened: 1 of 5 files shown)


src/ios/UAMessageViewController.m
```diff
@@ -2,26 +2,164 @@
...
+@end
+
 @implementation UAMessageViewController
 
 - (void) viewDidLoad {
     [super viewDidLoad];
 
+    self.airshipMessageViewController = [[UADefaultMessageCenterMessageViewController alloc]
+                                         initWithNibName:@"UADefaultMessageCenterMessageViewController"
+                                         bundle:[UAMessageCenterResources bundle]];
```
BrianBatchelder, line 17: This indentation looks weird.

src/ios/UAMessageViewController.m
```diff
@@ -2,26 +2,164 @@
...
+    self.airshipMessageViewController.delegate = self;
+
+
```
BrianBatchelder, line 20:
```suggestion
```

src/ios/UAirshipPlugin.m
```diff
@@ -719,32 +719,19 @@ - (void)displayInboxMessage:(CDVInvokedUrlCommand *)command {
...
-        UAMessageViewController *mvc = [[UAMessageViewController alloc] initWithNibName:@"UAMessageCenterMessageViewController"
-                                                                                 bundle:[UAMessageCenterResources bundle]];
+        UAMessageViewController *mvc = [[UAMessageViewController alloc] init];
+        mvc.modalTransitionStyle = UIModalTransitionStyleCrossDissolve;
+        mvc.modalPresentationStyle = UIModalPresentationFullScreen;
+        [[UIApplication sharedApplication].keyWindow.rootViewController presentViewController:mvc animated:YES completion:nil];
 
-        UINavigationController *navController =  [[UINavigationController alloc] initWithRootViewController:mvc];
+        // Load the message
+        [mvc loadMessageForID:[command.arguments firstObject]];
```
BrianBatchelder, line 730: Does this method still exist on `UAMessageViewController`? I think it may now be `displayMessageForID:`.

src/ios/UAMessageViewController.m
rlepinski: Its because usually you new line after a pram but this line is huge

PR #351, opened 2020-05-05
[MOBILE-1506] Update Android SDK 13.1.0

rlepinski:
### What do these changes do?
Updates latest Android SDK. Removes theme on manifest entries to make it easier to theme. The message center will now pick up app theme. App compat was not needed since we are not using it, but apps should be able to add it back if they want it.

Latest SDK!

Ran sample app

### DIFF
--- a/src/android/UAirshipPlugin.java
+++ b/src/android/UAirshipPlugin.java
@@ -29,9 +29,10 @@ import com.urbanairship.cordova.events.NotificationOpenedEvent;
 import com.urbanairship.google.PlayServicesUtils;
 import com.urbanairship.json.JsonMap;
 import com.urbanairship.json.JsonValue;
+import com.urbanairship.messagecenter.Inbox;
+import com.urbanairship.messagecenter.Message;
+import com.urbanairship.messagecenter.MessageCenter;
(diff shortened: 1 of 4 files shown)


src/android/UAirshipPlugin.java
```diff
@@ -121,9 +121,7 @@ public void onResume(boolean multitasking) {
 
     /**
      * To extend the plugin, add the actions to either {@link #AIRSHIP_ACTIONS} or {#link #GLOBAL_ACTIONS} and
-     * then define the method with the signature `void <CORDOVA_ACTION>(JSONArray data, final
-     * CallbackContext callbackContext)` and it will automatically be called. All methods will be
-     * ( executed in the ExecutorService. Any exceptions thrown by the actions are automatically caught
+     * All methods will be executed in the ExecutorService. Any exceptions thrown by the actions are automatically caught
```
BrianBatchelder, line 124: I think you lost something between these two lines.

rlepinski: Is no longer relevant

BrianBatchelder: But you're dangling an "and" at the end of the first line.

PR #363, opened 2020-09-28
Update Cordova module and Android and iOS SDK version

Ulrico972:
### What do these changes do?
Update cordova plugin as a patch (12.0.1).
Then Android SDK to 14.0.1 and iOS SDK to 14.1.2.

To keep cordova module up to date.

Build and run Android and iOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Version 12.0.1 - September 28, 2020
 
-Patch release updating iOS and SDK versions to 14.1.2 and 14.0.1, respectively.
+Patch release updating iOS and Android SDK versions to 14.1.2 and 14.0.1, respectively.
 
 ### iOS
 - Fixes a crash related to sending In-App Messages through push notifications.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,26 @@
 # Cordova Plugin Changelog
 
+## Version 12.0.1 - September 28, 2020
+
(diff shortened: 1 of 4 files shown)


rlepinski:
You should add a changelog so we can do a release from this PR

CHANGELOG.md
```diff
@@ -1,15 +1,27 @@
 # Cordova Plugin Changelog
 
+## Version 12.0.1 - September 28, 2020
+
+Patch release updating iOS and SDK versions to 14.1.2 and 14.0.1, respectively.
```
oristanovic, line 5: iOS and Android SDK?

Ulrico972: Right, thanks

rlepinski (approved):

package.json
```diff
@@ -50,5 +50,8 @@
   },
   "scripts": {
     "generate-docs": "jsdoc ./www/UrbanAirship.js -t ./node_modules/minami/ -r jsdoc_readme.md"
+  },
+  "dependencies": {
+    "cordova": "^10.0.0"
```
marc-scig, line 55: This seems like something worth mentioning elsewhere. Should we call it out in the README or CHANGELOG?

rlepinski: Oh, good catch. We should not have a direct dependency on cordova.

marc-scig: alternatively maybe we don't need it, or it should be a dev dependency?

Ulrico972: Oh, my mistake, I don't even remember added this.
Thank you

marc-scig (approved):

PR #365, opened 2020-11-29
FIX cordova spelling

aodysseos:
I've noticed this 'cordova' was misspelled 'cardova' in this file.

It seems that it is then been used at the following line:
`cordova.fireDocumentEvent(e.eventType, e.eventData)`

### DIFF
--- a/www/UrbanAirship.js
+++ b/www/UrbanAirship.js
@@ -1,6 +1,6 @@
 /* Copyright Urban Airship and Contributors */
 
-var cardova = require("cordova"),
+var cordova = require("cordova"),
     exec = require("cordova/exec"),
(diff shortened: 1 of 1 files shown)


rlepinski:
Thanks for the PR, we will try to get a release out asap

rlepinski:
The weird thing is the events work fine in our sample app. I assume its because `cordova` is loaded in the window as global. Did you have issues with events not firing? I might hold off on the release unless you are having issues.

aodysseos:
Hi, thank you for having a look at this. Yes, I couldn't get any events to fire after sending messages out. However, I was testing on an emulator. I will try on a device today and see if that works.

PR #369, opened 2021-02-11
[MOBILE-1847] Reorganize cordova plugins

Ulrico972:
### What do these changes do?
That reorganizes the cordova plugins as a mono repo and updates the scripts and GH workflows accordingly.

To have a better organization

Difficult to test all the things. I ran most of the scripts but I can't test the js doc uploading.


urbanairship-cordova/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "co-prompt": "^1.0.0",
+    "commander": "^2.20.3",
+    "fs": "0.0.1-security",
+    "jsdoc": "^3.6.6",
+    "minami": "1.2.3",
+    "semver": "^5.7.1",
+    "superagent": "^4.1.0"
+  },
+  "scripts": {
+    "generate-docs": "jsdoc ./www/UrbanAirship.js -t ./node_modules/minami/ -r ../jsdoc_readme.md"
```
marc-scig, line 52: When using yarn workspaces like this, all the installed node modules will end up in the root directory of the repo, and the .node_modules directories will be empty in the workspaces. So this will fail unless it's changed to look for minami another level back:

```
"generate-docs": "jsdoc ./www/UrbanAirship.js -t ../node_modules/minami/ -r ../jsdoc_readme.md"
```

marc-scig (commented):
Looking good, there's just a small change we need to make to the generate docs script. Also I think we'll need to update our github release workflow for publishing. Instead of `npm publish` in the root, we can run `yarn publish` in each workspace. react-native-module has a similar situation.

urbanairship-accengage-cordova/CHANGELOG
```diff
@@ -0,0 +1,8 @@
+# Airship Accengage Cordova Plugin Changelog
+
+## Version 11.0.1 - August 17, 2020
```
rlepinski, line 3: We will have a single version and changelog for all modules

urbanairship-accengage-cordova/src/android/build-extras.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+allprojects {
+    repositories {
+        google()
+        jcenter()
+    }
+}
+
+dependencies {
+    implementation 'com.urbanairship.android:urbanairship-accengage:13.3.2'
```
rlepinski, line 21: right version? I think we need 14.something

marc-scig (approved):
+1 with an update to the DEV_README that shows how to use the new yarn commands

rlepinski (approved):

PR #371, opened 2021-03-29
[MOBILE-2141] update SDK versions to 14.3.0

marc-scig:
Tested manually on a device/emulator with pushes and general poking around. I also made some changes to the update_version script which should make it easier for us to keep the version numbers up to date in the various modules.


Ulrico972 (approved):
Nice 👍

rlepinski (approved):

PR #383, opened 2022-01-07
[MOBILE-2800] fix push enabling on launch

Ulrico972:
### What do these changes do?
Set the push as enabled at launch if the config said so.

We used to set push as disabled at launch if the config enable_push_onlaunch is false.
That config field should only enable pushs at launch, not disable them.

Manual tests


rlepinski:
@Ulrico972 This is great but could you also grab my changes in the release PR?

rlepinski (approved):

Ulrico972:
> @Ulrico972 This is great but could you also grab my changes in the release PR?

Sure, I'll take care of it

jyaganeh (approved):

PR #405, opened 2022-12-05
[MOBILE-3381] Update to latest SDKs

Ulrico972:
### What do these changes do?
Update the SDKs and add support for custom domain

Keep up to date

Manual tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Cordova Plugin Changelog
 
-## Version 14.6.0 - December 5, 2022
+## Version 14.6.0 - December 7, 2022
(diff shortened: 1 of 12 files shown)


oristanovic (approved):

khmMouna (approved):

rlepinski:
@Ulrico972 Do we need to allow setting initial config from JS? we have a takeoff method right?

Ulrico972:
> @Ulrico972 Do we need to allow setting initial config from JS? we have a takeoff method right?

We could yeah. We actually only handle app and secret keys and cloud site from the current takeoff method lol.
I could add it and the rest of the keys if needed

rlepinski:
This is good enough for now

rlepinski (approved):

PR #407, opened 2023-01-10
Release 14.7.0

khmMouna:
### What do these changes do?
Prepare release 14.7.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Cordova Plugin Changelog
 
+## Version 14.7.0 - January 10, 2023
+Minor release adding support for styling message center. 
+
+### Changes
+- Add new methods `setMessageCenterStyleConfig` and `setMessageCenterTitle`.
+
 ## Version 14.6.0 - December 7, 2022
 Minor release updating Android SDK to 16.8.0 and iOS SDK to 16.10.6. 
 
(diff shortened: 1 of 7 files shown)


urbanairship-cordova/www/UrbanAirship.js
```diff
@@ -1113,6 +1114,32 @@ module.exports = {
...
 
+   /**
+   * Sets the message center style config file.
+   *
+   * @param {file} the message center style config file name.
+   * @param {function} [success] Success callback.
+   * @param {function(message)} [failure] Failure callback.
+   * @param {string} failure.message The error message.
+   */
+   setMessageCenterStyleConfig: function(file, success, failure) {
```
rlepinski, line 1125: What is the use case for this vs just pulling in a predefined file?

khmMouna: this will call:
`UAMessageCenter.shared.defaultUI.messageCenterStyle = [UAMessageCenterStyle styleWithContentsOfFile:fileName];` from the native SDK.
So it will override the predefined file if I well understood what you mean

rlepinski: But we already look for a default style class, so this is if they want to switch it at runtime. I would rather not do that here.

rlepinski: Maybe we should extend takeOff to allow these options?

khmMouna: It should be ok now. We can add the message center config style to the config file or use `messageCenterConfigStyle` by default.

Ulrico972 (approved):

PR #414, opened 2023-06-23
Release 14.10.1

rlepinski:
Fixes a crash on Android if takeOff is called with invalid config. We already validate the config on iOS so this is just an Android issue. Logging/error could be improved but I believe thats already the case in the proxy framework, so this is a quick fix for now until we move cordova over to the proxy.

### DIFF
--- a/urbanairship-cordova/src/android/PluginManager.java
+++ b/urbanairship-cordova/src/android/PluginManager.java
@@ -376,9 +376,12 @@ public class PluginManager {
         }
 
         try {
-            configOptions = builder.build();
+            AirshipConfigOptions config = builder.build();
+            config.validate();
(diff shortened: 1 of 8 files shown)


urbanairship-cordova/src/android/PluginManager.java
```diff
@@ -376,9 +376,12 @@ public AirshipConfigOptions getAirshipConfig() {
         }
 
         try {
-            configOptions = builder.build();
+            AirshipConfigOptions config = builder.build();
+            config.validate();
```
rlepinski, line 380: the magic

Ulrico972 (approved):

jyaganeh (approved):

PR #423, opened 2024-04-10
15.0.0 Issues

MaximBelov:
### What do these changes do?

1. Bug fix

2. Airship.contact.getNamedUserId should return an empty string "" instead "OK"
image
image

3. Airship.channel.getChannelId should return an empty string "" instead "OK"

Should fix a lot of errors

### DIFF
--- a/cordova-airship/types/index.d.ts
+++ b/cordova-airship/types/index.d.ts
@@ -1765,7 +1765,7 @@ export interface AirshipChannel {
      * @param error Error callback.
      */
     enableChannelCreation(
-        success?: (result: JsonValue) => void,
+        success?: () => void,
         error?: (err: string) => void
     ): void
 
--- a/cordova-airship/types/index.d.ts
+++ b/cordova-airship/types/index.d.ts
@@ -540,6 +540,7 @@ export namespace Android {
  * Enum of authorized Features.
  */
 export enum Feature {
+    All = 'all',
     InAppAutomation = 'in_app_automation',
     MessageCenter = 'message_center',
     Push = 'push',
@@ -551,7 +552,8 @@ export enum Feature {
 /**
  * All available features.
  */
-export const FEATURES_ALL = Object.values(Feature);
+// Error: TS1254: A const initializer in an ambient context must be a string or numeric literal or literal enum reference.
+// export const FEATURES_ALL = Object.values(Feature);
 
 /**
  * Custom event
@@ -956,7 +958,7 @@ export interface AirshipPush {
     /**
      * iOS only push methods.
      */
-    readonly iOS: AirshipPushIOS;
+    readonly ios: AirshipPushIOS;
 
 
     /**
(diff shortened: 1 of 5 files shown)


cordova-airship/www/Airship.js
```diff
@@ -386,6 +382,11 @@ airship.push.onNotificationResponse = function (callback) {
     return registerListener("airship.event.notification_response", callback)
 }
 
+airship.push.onNotificationReceived = function (callback) {
```
rlepinski, line 385: this is covered by onPushReceived

MaximBelov: It's not working as expected.

How to reproduce 
a. Cold start
1. Close application 
2. Send push notification via API
3. Tap on notification
4. Application Opened
5. Callback "onPushReceived" does not receive any event

MaximBelov: In this case, it is necessary to change MIGRATION.md
because it's not implemented here https://github.com/urbanairship/urbanairship-cordova/blob/main/cordova-airship/www/Airship.js

https://github.com/urbanairship/urbanairship-cordova/blob/main/MIGRATION.md

image

rlepinski: Yes, so this should be remove though before i can merge

cordova-airship/www/Airship.js
```diff
@@ -280,20 +276,20 @@ airship.channel.onChannelCreated = function (callback) {
...
     argscheck.checkArgs('FF', 'Airship.channel.enableChannelCreation', arguments)
     perform("channel#enableChannelCreation", null, success, failure)
 }
 
 // Contact
 
-airship.contact.getNamedUser = function (success, failure) {
-    argscheck.checkArgs('fF', 'Airship.contact.getNamedUser', arguments)
-    perform("contact#getNamedUser", null, success, failure)
+airship.contact.getNamedUserId = function (success, failure) {
```
rlepinski, line 286: Need to add back getNamedUser and deprecate it so its not a breaking change

cordova-airship/types/index.d.ts
```diff
@@ -1753,7 +1765,7 @@ export interface AirshipChannel {
      * @param error Error callback.
      */
     enableChannelCreation(
-        success?: () => void,
+        success?: (result: JsonValue) => void,
```
rlepinski, line 1768: this should be empty i believe

cordova-airship/www/Airship.js
MaximBelov: Could you explain what needs to be done?
I believe that there should be the same method everywhere

image

rlepinski: In Airship.js you need to define both `getNamedUser` and `getNamedUserId`. `getNamedUser` can call through to `getNamedUserId`. Add documentation to deprecate `getNamedUser`

rlepinski, line 286: I guess its currently broken so maybe what you have is fine

rlepinski:
I think what you currently have is great. Going to have another engineer on my team look over this as well. Thanks for the PR!

cordova-airship/www/Airship.js
rlepinski: Nvm, lets leave this since its currently broken

rlepinski:
Thanks!

PR #426, opened 2024-04-19
Add missing onBackgroundPushReceived method

MaximBelov:
### What do these changes do?

Add missing onBackgroundPushReceived method

Added method to Example/index,html

### DIFF
--- a/cordova-airship/src/android/AirshipCordova.kt
+++ b/cordova-airship/src/android/AirshipCordova.kt
@@ -2,13 +2,16 @@
 
 package com.urbanairship.cordova
 
+import android.annotation.SuppressLint
 import android.content.Context
+import android.content.Intent
 import android.os.Build
 import com.urbanairship.Autopilot
 import com.urbanairship.PendingResult
 import com.urbanairship.UALog
 import com.urbanairship.actions.ActionResult
 import com.urbanairship.android.framework.proxy.EventType
+import com.urbanairship.android.framework.proxy.Utils
 import com.urbanairship.android.framework.proxy.events.EventEmitter
 import com.urbanairship.android.framework.proxy.proxies.AirshipProxy
 import com.urbanairship.android.framework.proxy.proxies.FeatureFlagProxy
@@ -16,6 +19,10 @@ import com.urbanairship.json.JsonList
 import com.urbanairship.json.JsonMap
 import com.urbanairship.json.JsonSerializable
 import com.urbanairship.json.JsonValue
+import com.urbanairship.json.jsonMapOf
+import com.urbanairship.push.NotificationInfo
+import com.urbanairship.push.PushManager
+import com.urbanairship.push.PushMessage
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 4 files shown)


cordova-airship/src/android/AirshipCordova.kt
```diff
@@ -87,6 +95,33 @@ class AirshipCordova : CordovaPlugin() {
                 notifyPendingEvents()
             }
         }
+
+        cordova.getThreadPool().execute {
+            try {
+                notifyListeners(
+                    EventType.BACKGROUND_NOTIFICATION_RESPONSE_RECEIVED,
```
rlepinski, line 102: The problem is the cordova context is setup in the activity not the application. I dont think this will help but maybe I am missing something

cordova-airship/src/android/AirshipCordova.kt
```diff
@@ -97,6 +132,34 @@ class AirshipCordova : CordovaPlugin() {
...
+                    notificationInfo.message,
+                ),
+            ).pluginResult()
+            if (listeners?.isNotEmpty() == true) {
+                listeners.forEach { listener ->
+                    pluginResult.keepCallback = true
+                    listener.callbackContext.sendPluginResult(pluginResult)
+                }
+            } else {
+                this.notificationStack.getOrPut(eventType) { mutableListOf() }.add(notificationInfo)
```
rlepinski, line 160: The proxy already queues the events and waits an active listener. All events should send once you have a listener on the cordova side

cordova-airship/src/android/AirshipCordova.kt
MaximBelov: you're right
Thank you

cordova-airship/www/Airship.js
```diff
@@ -377,6 +377,11 @@ airship.push.onPushReceived = function (callback) {
     return registerListener("airship.event.push_received", callback)
 }
 
+airship.push.onBackgroundPushReceived = function (callback) {
```
rlepinski, line 380: This is only different on Android because on the other frameworks you have to do some special hoops to get background push working. On cordova I am not sure I see a point of separating them.  What if we have both `airship.event.background_push_received` and `airship.event.push_received` go to the same listener?

MaximBelov: depending on the state of the application in the background only one will work 

`airship.event.push_received` - foreground, ios and android
`airship.event.notification_response` - background, ios only
`airship.event.background_push_received`  - background android only -  background or cold start 

my initial testing with positive results
it works correctly without any problems

rlepinski: But do you need it separated? The event map, I think we can just map the background event to airship.event.push_received on Android to have a single onPushReceived

MaximBelov: we can combine this into one event listener, but we must add new properties to the response to understand exactly what type of event received and where it came from

rlepinski: You need this exposed in the app?

rlepinski: Looks like you will get a foreground on the response but not currently the push received event. We could add it though

rlepinski: Adds isForeground and orders events - https://github.com/urbanairship/airship-mobile-framework-proxy/pull/39

Going to open a new PR that combines pushReceived and notificationResponses into the same listener

rlepinski:
https://github.com/urbanairship/urbanairship-cordova/pull/428 still need the proxy to be reviewed + released before the event order is correct and we get the `isForeground` flag in the events

MaximBelov:
Fixed
https://github.com/urbanairship/urbanairship-cordova/pull/428

PR #427, opened 2024-04-19
Android null result

MaximBelov:
### What do these changes do?

Standardise null result on android and ios

Should return a "null" result on Android  instead of "OK"
android – AirshipCordova kt  AirshipCordova app main  2024-04-19 16-04-55

Run demo app

android – AirshipCordova kt  AirshipCordova app main  2024-04-19 16-07-01

iPhone
android – AirshipCordova kt  AirshipCordova app main  2024-04-19 16-07-01

### DIFF
--- a/cordova-airship/src/android/AirshipCordova.kt
+++ b/cordova-airship/src/android/AirshipCordova.kt
@@ -396,7 +396,7 @@ internal fun JsonSerializable.pluginResult(): PluginResult {
     val json = this.toJsonValue()
 
     return when {
-        json.isNull -> PluginResult(PluginResult.Status.OK)
+        json.isNull -> PluginResult(PluginResult.Status.OK, json.string)
         json.isString -> PluginResult(PluginResult.Status.OK, json.requireString())
         json.isBoolean -> PluginResult(PluginResult.Status.OK, json.requireBoolean())
         json.isInteger -> PluginResult(PluginResult.Status.OK, json.getInt(0))
--- a/cordova-airship/src/android/AirshipCordova.kt
+++ b/cordova-airship/src/android/AirshipCordova.kt
@@ -396,7 +396,7 @@ internal fun JsonSerializable.pluginResult(): PluginResult {
     val json = this.toJsonValue()
 
     return when {
-        json.isNull -> PluginResult(PluginResult.Status.OK, json.string)
+        json.isNull -> PluginResult(PluginResult.Status.OK, "")
         json.isString -> PluginResult(PluginResult.Status.OK, json.requireString())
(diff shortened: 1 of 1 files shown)


cordova-airship/src/android/AirshipCordova.kt
```diff
@@ -396,7 +396,7 @@ internal fun JsonSerializable.pluginResult(): PluginResult {
     val json = this.toJsonValue()
 
     return when {
-        json.isNull -> PluginResult(PluginResult.Status.OK)
+        json.isNull -> PluginResult(PluginResult.Status.OK, json.string)
```
rlepinski, line 399: instead of json.string should it just be `null` ?

MaximBelov: null is not supported by cordova 
Screenshot 2024-04-19 at 17 34 42

rlepinski: oh, this is literally passing "null"

rlepinski: Looks like we use to do an empty string for null - https://github.com/urbanairship/urbanairship-cordova/blob/14.11.0/urbanairship-cordova/src/android/UAirshipPlugin.java#L794

MaximBelov: Yes, an empty string is fine too

rlepinski: I am sure your change is good but let me double check the difference between iOS and Android

rlepinski: There is also https://documentation.opencms.org/javadoc/core/org/opencms/json/JSONObject.Null.html that we can try

rlepinski: It looks like it does support null - https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/PluginResult.java#L42C46-L42C63

try `(String)null`

rlepinski:
https://github.com/urbanairship/urbanairship-cordova/pull/428/files

MaximBelov:
Fixed
https://github.com/urbanairship/urbanairship-cordova/pull/428/files#diff-9a808c9c8deee161f9c5b48f6bf9d6399a4d0604ca921a40ad7d59ffeb7f6004R385

MaximBelov:
Fixed
https://github.com/urbanairship/urbanairship-cordova/pull/428/files#diff-9a808c9c8deee161f9c5b48f6bf9d6399a4d0604ca921a40ad7d59ffeb7f6004R385

